PR #484, opened 2023-08-01
Add batching support to concepts.

nsthorat:

- Adds a batch_utils.py which has flatten, chunks, and flat_batch_compute, which allows you to batch call f() on nested object (only list[list[Any]])
- Add a /compute endpoint for arbitrary signals.

Details to make this work:
- Fixing circular deps by moving lilac_span, lilac_embedding, SpanVector to schema.py

image


dsmilkov (approved):
very nice cleanup!

PR #506, opened 2023-08-04
Add a signal preview page.

nsthorat:
- Add a signal preview page which lets you paste text and preview a signal's output.
- Change the concept preview to use the schema endpoint for the concept scorer.

NOTE: This doesn't share logic with the concept preview because they are different enough.

https://lilacai-nikhil-staging.hf.space/signals#lang_detection

image


dsmilkov (approved):
yay!

PR #508, opened 2023-08-07
Remove implicit negatives and update the concept creation UI to sample from dataset.

dsmilkov:
https://github.com/lilacai/lilac/assets/2294279/f6aa81b6-b61a-4325-b904-e3f4e19ce1bb


lilac/data/dataset.py
```diff
@@ -282,11 +281,6 @@ def update_settings(self, settings: DatasetSettings) -> None:
     """Update the settings for the dataset."""
     pass
 
-  @abc.abstractmethod
-  def get_vector_db_index(self, embedding: str, path: PathTuple) -> VectorDBIndex:
```
nsthorat, line 286: in my PR I'm using this to determine whether an embedding is computed on a column already from a base dataset class

maybe I should just be reading the manifest?

nsthorat (approved):

lilac/data/dataset.py
dsmilkov: Yup, let's read the manifest. signal manifests now have optional `vector_store`, which is only set when there is an embedding computed.

PR #510, opened 2023-08-07
Add a load script which reads a yml file that configures a lilac instance.

nsthorat:


lilac/load.py
```diff
@@ -0,0 +1,209 @@
...
+    if not embedding_field:
+      compute_embedding = True
+
+  if compute_embedding:
+    dataset.compute_embedding(embedding_config.embedding, embedding_config.path, task_step_id)
+
+  gc.collect()
+
+
+if __name__ == '__main__':
```
dsmilkov, line 208: can you register this as a subcommand in `cli.py`. So users can do `lilac load ...` in addition to `lilac start` and `lilac version`.

dsmilkov (approved):

lilac/load.py
```diff
@@ -0,0 +1,209 @@
...
+import pathlib
+import shutil
+
+import click
+import dask
+import psutil
+import yaml
+from distributed import Client
+
+from lilac.config import Config, EmbeddingConfig, SignalConfig
```
dsmilkov, line 22: use relative imports here

lilac/config.py
```diff
@@ -1,63 +1,81 @@
...
-
-
-# Initialize the environment at import time.
-_init_env()
+"""Configurations for a dataset run."""
+from typing import Optional
+
+from pydantic import BaseModel, validator
+
+from lilac.sources.source_registry import resolve_source
```
dsmilkov, line 6: use relative import

lilac/load.py
nsthorat: done

lilac/load.py
nsthorat: done

lilac/config.py
nsthorat: done

PR #513, opened 2023-08-08
Add a deploy_demo script which deploys our public demo.

nsthorat:
https://huggingface.co/spaces/lilacai/lilac

### DIFF
--- a/demo.yml
+++ b/demo.yml
@@ -4,7 +4,7 @@ datasets:
     settings:
       ui:
         media_paths: ['text']
-        preferred_embedding: 'gte-small'
+      preferred_embedding: 'gte-small'
     source:
       source_name: huggingface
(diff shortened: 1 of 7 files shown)


demo.yml
```diff
@@ -4,7 +4,7 @@ datasets:
     settings:
       ui:
         media_paths: ['text']
-        preferred_embedding: 'gte-small'
+      preferred_embedding: 'gte-small'
```
dsmilkov, line 7: can we validate a yaml? i guess via pydantic it happens automatically?

dsmilkov (approved):

demo.yml
nsthorat: added to all the configs:

```
  class Config:
    extra = Extra.forbid
```

PR #518, opened 2023-08-08
Keep the config.yml file in sync when updating settings, running signals and embeddings.

nsthorat:
Python:
- settings.json is now gone, everything happens through config.yml
- The public API for update_settings() and settings() remains. It's an impl detail it's in config.
- Add a public API for data.config() which returns a DatasetConfig object.
- Relax the config.yml for load() to allow both a Config (multi-dataset) or DatasetConfig (single-dataset).

UI:
- Show a modal dialog from the (i) button next to the dataset name which shows the config yaml used to generate this dataset.

NOTE: This PR should be backwards compatible. If there is an old settings.json file, the new config.yml will be written and reflect the old settings. HOWEVER: the config.yml will not reflect embeddings or signals that were run.

Modal:
image


pyproject.toml
```diff
@@ -137,7 +138,6 @@ types-tqdm = "^4.65.0.0"
 watchdog = { extras = ["watchmedo"], version = "^3.0.0" }
 wheel = "^0.37.1"
 yapf = "^0.32.0"
-pyyaml = "^6.0.1"
 
 [tool.poetry.scripts]
 deploy-hf = "scripts.deploy_hf:main"
```
dsmilkov, line 143: remove `deploy-hf` from scripts?

lilac/config.py
```diff
@@ -28,6 +39,32 @@ def parse_signal(cls, signal: dict) -> Signal:
     """Parse a signal to its specific subclass instance."""
     return resolve_signal(signal)
 
+  def dict(
```
dsmilkov, line 42: i feel this is cleaner done via customizing the yaml/json serializer

lilac/config.py
```diff
@@ -28,6 +39,32 @@ def parse_signal(cls, signal: dict) -> Signal:
...
+    exclude: Optional[Union['AbstractSetIntStr', 'MappingIntStrAny']] = None,
+    by_alias: bool = False,
+    skip_defaults: Optional[bool] = None,
+    exclude_unset: bool = False,
+    exclude_defaults: bool = False,
+    exclude_none: bool = False,
+  ) -> dict[str, Any]:
+    """Override the default dict method to simplify the path tuples.
+
+    This is required to remove the python-specific tuple dump in the yaml file.
```
dsmilkov, line 55: do tuples cause problems in yaml? or is it purely ergonomics?

lilac/data/dataset_duckdb.py
```diff
@@ -241,22 +260,61 @@ def manifest(self) -> DatasetManifest:
       latest_mtime_micro_sec = int(latest_mtime * 1e6)
       return self._recompute_joint_table(latest_mtime_micro_sec)
 
+  def _update_config(self,
```
dsmilkov, line 263: i feel like it would be awesome if we can avoid an "update" method, and have the client send a entire new settings to replace the old one. update methods tend to get very complex over time and often we fail to correctly perform an update

dsmilkov (approved):
nice!

PR #520, opened 2023-08-08
Add popular datasets to the demo.

nsthorat:
https://huggingface.co/datasets?sort=downloads

These are all relatively small datasets (<1M)

image


dsmilkov (approved):
great!

PR #525, opened 2023-08-10
Add guides skeleton.

nsthorat:
Just adding the top-level structure for review before going deep on each.

image


dsmilkov (approved):
awesome!

PR #534, opened 2023-08-11
Move lilac concepts into the pip package.

nsthorat:
Lilac concepts are now in lilac/concepts. This is done by using importlib resources to find the files.

Other changes:

- .pkl model files now live in DATA_PATH/.cache/lilac/concept. This allows us to choose to sync these files.
- deploy_hf has a new bit skip_cache which can skip uploading local .pkl files.
- Remove the top-level __init__.py. This was there for scripts to import lilac. They can now import via public API.
- Add simple CLI for `lilac concepts` so we can make sure concepts are in the final package.


.gitignore
```diff
@@ -17,10 +17,7 @@ node_modules/
 dist/
 
 # Data to ignore.
-data/datasets
-data/.*
-data/concept/*
-!data/concept/lilac
+data/
```
dsmilkov, line 20: just a future thought: if we default LILAC_DATA_PATH to ~/data (homedir), then we can remove this data/ in gitignore since that folder no longer exists on github

dsmilkov (approved):
amazing

.gitignore
nsthorat: That's a great idea, it's probably not great that it's relative to the package, since it will be in a strange location when someone pip installs

PR #551, opened 2023-08-15
Add medical_dialogue, clean up demo.

nsthorat:
- Remove spacy_ner from demo datasets
- Add medical_dialog
- Add the_movies_dataset (fix issues with public bucket.. we should probably make a kaggle source)


demo.yml
```diff
@@ -55,34 +55,33 @@ datasets:
...
+  - name: databricks-dolly-15k-curated-en
+    namespace: lilac
+    settings:
+      ui:
+        media_paths:
+          - original-instruction
+          - original-context
+          - original-response
+          - - new-instruction
+            - value
```
dsmilkov, line 67: i think you can make this an array, like in line 80

scripts/deploy_demo.py
```diff
@@ -68,11 +68,6 @@ def deploy_demo(overwrite: bool, skip_sync: bool, skip_load: bool, skip_build: b
   if not skip_load:
     load(DEMO_DATA_DIR, DEMO_CONFIG_PATH, overwrite)
 
-  # Copy lilac concepts to the demo data dir from the default data_path: 'data'.
```
dsmilkov, line 71: is this gone because deploy_hf does it?

dsmilkov (approved):

demo.yml
nsthorat: done

scripts/deploy_demo.py
nsthorat: yeah, it happens when we push source code now

PR #569, opened 2023-08-16
Update questions concept with more examples. Fix concept UI bugs.

nsthorat:
Add lots of labels for questions. I can't get the gte-small past 'Good':

image

Labeler:
- Show a warning message when there's no embedding:
image

- Fix the link to the dataset in the labeler

- Fix labeling issue where if a label takes a while to commit, we don't hang the UI.

Concept view:
- Separate metrics to its own component so individual metrics can update separately.

### DIFF
--- a/web/blueprint/src/lib/components/concepts/ConceptMetrics.svelte
+++ b/web/blueprint/src/lib/components/concepts/ConceptMetrics.svelte
@@ -36,9 +36,16 @@
       {scoreToText[$model.data.metrics.overall]}
     </div>
   {:else}
+    {@const createModelIfNotExists = true}
     <Button
       icon={$modelMutation.isLoading ? InlineLoading : Chip}
-      on:click={() => $modelMutation.mutate([concept.namespace, concept.concept_name, embedding])}
+      on:click={() =>
+        $modelMutation.mutate([
+          concept.namespace,
+          concept.concept_name,
+          embedding,
+          createModelIfNotExists
+        ])}
       class="w-28 text-3xl"
     >
       Compute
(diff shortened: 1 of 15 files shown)


web/blueprint/src/lib/components/concepts/labeler/ConceptLabeler.svelte
```diff
@@ -121,17 +120,28 @@
     path = deserializePath(val);
   }
 
+  $: datasetViewStore =
```
dsmilkov, line 123: should we move the creation of `datasetViewStore` inside `openDataset()` since we only create a temporary state for the redirect, and we want to avoid a situation where we accumulate state (e.g. accumulate 2 searches via addSearch)

dsmilkov (approved):

web/blueprint/src/lib/components/concepts/ConceptMetrics.svelte
```diff
@@ -0,0 +1,47 @@
+<script lang="ts">
```
dsmilkov, line 1: i just checked out your branch. There is one important regression. At head, If we never instantiated a model for a specific embedding (e.g. we never instantiated toxicity/cohere), opening the page for the toxicity model should not cause a creation of that model (e.g. we don't have a key for cohere). Only upon clicking "compute" did we instantiate a model for that embedding

dsmilkov (approved):

web/blueprint/src/lib/components/concepts/labeler/ConceptLabeler.svelte
nsthorat: Unfortunately you can't create a store inside the function -- it has to be defined top-level.

web/blueprint/src/lib/components/concepts/ConceptMetrics.svelte
nsthorat: done!

PR #574, opened 2023-08-17
Add export dataset, dataset signals, signals, custom signals docs

nsthorat:
https://lilacml.com/


docs/datasets/dataset_export.md
```diff
@@ -1 +1,67 @@
...
+`text.language_detection`:
+
+```python
+dataset = ll.get_dataset('local', 'imdb')
+
+# Select 'text' and 'text.language_detection'.
+# NOTE: you can select lang detection with 'text.lang_detection' or ('text', 'language_detection')
+results = dataset.select_rows(['text', ('text', 'lang_detection'), 'text'])
+
+# Print the first result.
```
dsmilkov, line 35: can you mention that results is iterable instead of a list

dsmilkov (approved):

docs/datasets/dataset_export.md
nsthorat: done

PR #580, opened 2023-08-21
Add Google Analytics conditionally

dsmilkov:


web/blueprint/.env
```diff
@@ -0,0 +1 @@
+PUBLIC_HF_ANALYTICS=0 # Analytics disabled by default.
```
nsthorat, line 1: can you put a comment at the top here about the lifecycle of this flag (e.g. it's not visible from py)

nsthorat (approved):

web/blueprint/.env
dsmilkov: done

PR #596, opened 2023-08-25
Add filter options along source and signal fields

dsmilkov:
Add more options in the search menu to make filtering/sorting easier:

- Filter signal fields by EXISTS (e.g. emails, ip addresses, etc)
- Sorting both source and signals fields

https://github.com/lilacai/lilac/assets/2294279/b14ef225-2c50-4c52-88cf-4eabcbb03f32


nsthorat (approved):
Nice work!

Few high-level suggestions:
image

- Can we separate the concepts from signals from metadata visually? Maybe metadata is gray, signals are blue, and concepts are green? Or subheaders for them? It feels like a lot of stuff is in there now
- Can the description have a little more left-padding? local/hacking  description is quite close to the labe;

dsmilkov:
done and done

PR #603, opened 2023-08-26
Simplify the search bar

dsmilkov:
DEMO: https://huggingface.co/spaces/lilacai/daniel_staging

Before
Screenshot 2023-08-26 at 11 48 46 AM

After:

https://github.com/lilacai/lilac/assets/2294279/64636895-3638-4ea1-a0ce-64cdba69529d

### DIFF
--- a/web/blueprint/src/lib/view_utils.ts
+++ b/web/blueprint/src/lib/view_utils.ts
@@ -271,12 +271,15 @@ export function getDefaultSearchPath(
   mediaPaths: Path[]
 ): Path | undefined {
   if (datasetStore.stats == null || datasetStore.stats.length === 0) {
-    return mediaPaths[0];
+    return undefined;
   }
   // The longest visible path that has an embedding is auto-selected.
   const mediaPathsStats = datasetStore.stats.filter(s =>
     mediaPaths.some(p => pathIsEqual(s.path, p))
   );
+  if (mediaPathsStats.length === 0) {
+    return undefined;
+  }
   let paths = mediaPathsStats.map(stat => {
     return {
       path: stat.path,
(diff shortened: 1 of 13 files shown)


web/blueprint/src/lib/view_utils.ts
```diff
@@ -276,28 +266,29 @@ export function getSearches(store: DatasetViewState, path?: Path | null): Search
   return (store.query.searches || []).filter(s => pathIsEqual(s.path, path));
 }
 
-function getDefaultSearchPath(datasetStore: DatasetState): Path | null {
+export function getDefaultSearchPath(
```
nsthorat, line 269: later we should make this part of the settings

nsthorat (approved):
Amazing stuff, much nicer.

High level comments:
- Can you round the left of the dropdown and right of the searchbox (rounded-l and rounded-r tailwind classes) to match the rest of our design?
- Can we use a different icon when the index has been computed, or something to separate the overloaded chip icon (which means "it has been computed" in the left dropdown and "it hasn't been computed, compute" in the right dropdown?) Ideally it's like a chip+check but not sure if that icon exists. A color might do it
- I get a funky R/L scroll:
image
- I can check the same filter twice (adding the same filter twice). And then if I remove it by clicking the X on the tag, it removes both:
image
- Clicking on near_dip.cluster_id does nothing (same with num_characters, readability, TTR, frac_non_ascii):
image
- When I add a filter for has emails for two columns, we dont show the column in the pill, we may want to group the pills by the column so we dont show the path for every pill:
image

web/blueprint/src/lib/view_utils.ts
dsmilkov: ack

dsmilkov:
great comments -- addressed all but 2

> Clicking on near_dip.cluster_id does nothing (same with num_characters, readability, TTR, frac_non_ascii):

it applies a sort DESC, might not be obvious, open to ideas on how to clarify that

> When I add a filter for has emails for two columns, we dont show the column in the pill

Issue #607, opened 2023-08-28
Opening lilac on local network


Orsucciu:
Hello,

I am trying to use lilac on my remote server.

I tried starting it from cli (lilac start) and then joining it through its ip adress, as well as using the python console to run lilac.server_start with options set to 0.0.0.0 as well as the ip of the server. Neither solutions give me errors in the console, but the server is unreachable.

I am using debian, with python 3.11 if that's relevant.

Thanks

nsthorat:
Hi @Orsucciu, can you try explicitly passing `--host 0.0.0.0 --port 3000` (or whatever port you want)?

dsmilkov:
The default Lilac port is `5432` so I just want to double check that you tried `remote_ip:5432`?

Alx-AI:
If you're using default settings, while it may say ``` Uvicorn running on http://0.0.0.0:5432``` navigating to http://localhost:5432/ instead might solve your issue

Orsucciu:
Hi, i just ran `lilac start --host 0.0.0.0 --port 4321` but when i get the ip directly (http://ip:port), i still get no answer (maybe this is a config problem, and not a lilac one tho!). The console doesn't send any errors either, my last lilac messages are 
`INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:4321 (Press CTRL+C to quit)`

nsthorat:
This should now be fixed at https://github.com/lilacai/lilac/releases/tag/v0.0.15, we now default to 127.0.0.1 so it should open to local 127.0.0.1 without the 0.0.0.0 interface.

Let me know if you have issues after upgrading

PR #609, opened 2023-08-28
Use pip for the HuggingFace demos.

nsthorat:
- Adds the flag --use_pip for the deploy_hf, which is the default for deploy_demo.
- Also default the host to `127.0.0.1`, which is the uvicorn default. The old default, `0.0.0.0` causes lots of issues with the server not being visible: https://github.com/lilacai/lilac/issues/607
- Build the webserver typescript before publishing on pip (0.0.13 has old typescript because we forgot to build)
- Move the analytics environment variable to a runtime flag. This is required because we use the public pip package from huggingface now.

New demo is deployed (after 0.0.14 push): https://huggingface.co/spaces/lilacai/lilac

### DIFF
--- a/lilac/cli.py
+++ b/lilac/cli.py
@@ -12,7 +12,7 @@ from .server import start_server
 @click.option(
   '--host',
   help='The host address where the web server will listen to.',
-  default='0.0.0.0',
+  default='127.0.0.1',
   type=str)
 @click.option('--port', help='The port number of the web-server', type=int, default=5432)
 def start(host: str, port: int) -> None:
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,19 +13,17 @@ ENV HOME=/home/user \
 # Set the working directory in the container.
 WORKDIR $HOME/app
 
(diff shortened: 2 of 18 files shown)


lilac/cli.py
```diff
@@ -12,7 +12,7 @@
 @click.option(
   '--host',
   help='The host address where the web server will listen to.',
-  default='0.0.0.0',
+  default='127.0.0.1',
```
dsmilkov, line 15: not sure this is correct, since the documentation explicitly says to use `--host 0.0.0.0` to make the application available on your local network. `127.0.0.1` will constrain it to only be visible on a single machine

https://www.uvicorn.org/settings/#socket-binding

dsmilkov (approved):
awesome!

lilac/cli.py
nsthorat: I think 127.0.0.1 is the right default:

1) it’s the uvicorn default
2) not all machines automatically make that available locally without a forward so users are confused
3) you don’t wanna accidentally make something private visible

PR #621, opened 2023-08-30
Add a project_path to lilac_start.

nsthorat:
`lilac start ~/my_project` will now start a project in that folder, create a root lilac.yml, and start the webserver. When the server boots up, we will call `load` which will show the tasks to load the datasets from the lilac.yml.

Impl details:
- When passing a project path, we will check for lilac.yml. If it doesn't exist, create it. This is now the project directory.

NOTE: The huggingface demos don't yet call the load on boot because there is no lilac.yml file. This is because we call `gunicorn server:app` directly, bypassing our public API of start_server. I'll clean this up in a follow up.

Nothing changes in demo: https://huggingface.co/spaces/lilacai/nikhil_staging


lilac/project.py
```diff
@@ -0,0 +1,47 @@
...
+  """Creates an empty lilac project."""
+  with open(os.path.join(project_path, PROJECT_CONFIG_FILENAME), 'w') as f:
+    f.write('# Lilac project config. See https://lilacml.com/api_reference/index.html#lilac.Config '
+            'for details.\n\n' +
+            to_yaml(Config(datasets=[]).dict(exclude_defaults=True, exclude_none=True)))
+
+
+def create_project_and_set_env(project_path_arg: str) -> None:
+  """Creates a Lilac project if it doesn't exist and set the environment variable."""
+  project_path = project_path_from_args(project_path_arg)
```
dsmilkov, line 40: should we do project_path = os.path.expanduser(project_path) so stuff like ~ is replaced with homedir

dsmilkov (approved):

lilac/project.py
nsthorat: Good idea, done in the helper util.

PR #625, opened 2023-08-31
Allow results of tanstack queries to contain non-serializable data

dsmilkov:
This fixes the P0 bug and protects us in the future from other query results that have pointers.


nsthorat (approved):
Ah nice!!

Issue #628, opened 2023-08-31
Vector store errors with no-key in id_to_span when flipping concept search sort order


nsthorat:
```
Traceback (most recent call last):
File "/home/user/.local/lib/python3.9/site-packages/lilac/router_utils.py", line 24, in custom_route_handler return await original_route_handler(request)
File "/home/user/.local/lib/python3.9/site-packages/fastapi/routing.py", line 241, in app raw_response = await run_endpoint_function(
File "/home/user/.local/lib/python3.9/site-packages/fastapi/routing.py", line 169, in run_endpoint_function return await run_in_threadpool(dependant.call, **values)
File "/home/user/.local/lib/python3.9/site-packages/starlette/concurrency.py", line 41, in run_in_threadpool return await anyio.to_thread.run_sync(func, *args)
File "/home/user/.local/lib/python3.9/site-packages/anyio/to_thread.py", line 33, in run_sync return await get_async_backend().run_sync_in_worker_thread(
File "/home/user/.local/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 2106, in run_sync_in_worker_thread return await future
File "/home/user/.local/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 833, in run result = context.run(func, *args) File "/home/user/.local/lib/python3.9/site-packages/lilac/router_dataset.py", line 230, in select_rows res = dataset.select_rows(
File "/home/user/.local/lib/python3.9/site-packages/lilac/data/dataset_duckdb.py", line 1048, in select_rows df[signal_column] = deep_unflatten(signal_out, input)
File "/home/user/.local/lib/python3.9/site-packages/lilac/batch_utils.py", line 51, in deep_unflatten return cast(list, _deep_unflatten(iter(flat_input), original_input, is_primitive_predicate))
File "/home/user/.local/lib/python3.9/site-packages/lilac/batch_utils.py", line 44, in _deep_unflatten return [_deep_unflatten(flat_input, orig_elem, is_primitive_predicate) for orig_elem in values]
File "/home/user/.local/lib/python3.9/site-packages/lilac/batch_utils.py", line 44, in <listcomp> return [_deep_unflatten(flat_input, orig_elem, is_primitive_predicate) for orig_elem in values]
File "/home/user/.local/lib/python3.9/site-packages/lilac/batch_utils.py", line 37, in _deep_unflatten return next(flat_input)
[...]

Issue #608, opened 2023-08-28
UnicodeEncodeError when using signals (even PII)


Alx-AI:
While running through the Quick Start guide, when running the PII detection signal, I get encode errors even while using the suggested OpenOrca dataset (on Windows 11 Python 3.10.4 FWIW):

P.S. Thank you for open sourcing such an incredible tool! 

```
Route error: http://localhost:5432/api/v1/concepts/
not enough values to unpack (expected 2, got 1)
Traceback (most recent call last):
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\lilac\router_utils.py", line 24, in custom_route_handler
    return await original_route_handler(request)
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\fastapi\routing.py", line 241, in app
    raw_response = await run_endpoint_function(
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\fastapi\routing.py", line 169, in run_endpoint_function
    return await run_in_threadpool(dependant.call, **values)
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\starlette\concurrency.py", line 41, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\anyio\to_thread.py", line 33, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\anyio\_backends\_asyncio.py", line 877, in run_sync_in_worker_thread
    return await future
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\anyio\_backends\_asyncio.py", line 807, in run
    result = context.run(func, *args)
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\lilac\router_concept.py", line 32, in get_concepts
    return DISK_CONCEPT_DB.list(user)
  File "C:\Users\ajduf\anaconda3\envs\lilac\lib\site-packages\lilac\concepts\db_concept.py", line 379, in list
    namespace, name = root.split('/')[-2:]
ValueError: not enough values to unpack (expected 2, got 1)

Route error: http://localhost:5432/api/v1/concepts/
not enough values to unpack (expected 2, got 1)
Traceback (most recent call last):
[...]

PR #629, opened 2023-09-01
Ds langsmith

dsmilkov:
Fixes https://github.com/lilacai/lilac/issues/628, https://github.com/lilacai/lilac/issues/608

- Fix the vector store key not found error when the embedding is sparse
- Fix the unicode encoding error for pii
- Ignore media fields in metadata

### DIFF
--- a/web/blueprint/src/lib/components/datasetView/RowItemMetadata.svelte
+++ b/web/blueprint/src/lib/components/datasetView/RowItemMetadata.svelte
@@ -7,6 +7,7 @@
     getField,
     isSignalRootField,
     listValueNodes,
+    pathIsEqual,
     serializePath,
     valueAtPath,
     type DataTypeCasted,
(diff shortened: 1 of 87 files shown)


web/blueprint/src/lib/components/datasetView/RowItemMetadata.svelte
```diff
@@ -36,7 +38,10 @@
     formattedValue?: string | null;
   }
   function makeRows(row: LilacValueNode): MetadataRow[] {
-    const valueNodes = listValueNodes(row).filter(item => isItemVisible(item, visibleFields));
+    const valueNodes = listValueNodes(row)
+      .filter(item => isItemVisible(item, visibleFields))
+      // Filter out media fields.
+      .filter(item => !mediaFields.some(mf => pathIsEqual(mf.path, L.field(item)?.path)));
```
nsthorat, line 44: just double checking that when you select a signal that's a child of a media field it still shows up

nsthorat (approved):

web/blueprint/src/lib/components/datasetView/RowItemMetadata.svelte
dsmilkov: good point. it does since we do exact path equals instead of path contains

PR #631, opened 2023-09-01
Update the global project config during changes.

nsthorat:
High-level: lilac.yml now will be updated any time an action is performed on a dataset.

Details:
- Datasets call into methods in `project` which will write to lilac.yml.
- The old config.yml is now gone, however for backwards compat I made sure to have lilac.yml update from dataset configs if they exist.
- `lilac.yml` is written *before* computation. The manifest will reflect the post-computation state. This means that when the server boots up, it will try to call load on the fresh lilac.yml.

### DIFF
--- a/lilac/data/dataset.py
+++ b/lilac/data/dataset.py
@@ -540,7 +540,11 @@ def default_settings(dataset: Dataset) -> DatasetSettings:
 
 
 def dataset_config_from_manifest(manifest: DatasetManifest) -> DatasetConfig:
-  """Computes a DatasetConfig from a manifest. Used for diffing the results to the config."""
+  """Computes a DatasetConfig from a manifest.
+
+  NOTE: This is only used for backwards compatibility. Once we remove the back-compat logic, this
+  method can be removed.
+  """
   all_fields = manifest.data_schema.all_fields
 
   all_signals = [
--- a/lilac/data/dataset.py
+++ b/lilac/data/dataset.py
@@ -21,13 +21,23 @@ from pydantic import (
 from typing_extensions import TypeAlias
 
 from ..auth import UserInfo
-from ..config import DatasetConfig, DatasetSettings, DatasetUISettings
+from ..config import (
+  DatasetConfig,
+  DatasetSettings,
+  DatasetUISettings,
+  EmbeddingConfig,
+  SignalConfig,
+  get_dataset_config,
+)
+from ..env import data_path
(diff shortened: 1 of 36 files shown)


lilac/data/dataset.py
```diff
@@ -497,6 +539,33 @@ def default_settings(dataset: Dataset) -> DatasetSettings:
   return DatasetSettings(ui=DatasetUISettings(media_paths=media_paths))
 
 
+def dataset_config_from_manifest(manifest: DatasetManifest) -> DatasetConfig:
```
dsmilkov, line 542: would this method go away once we remove back-compat? might be worth adding a comment if so

lilac/utils.py
```diff
@@ -283,9 +283,17 @@ def pretty_timedelta(delta: timedelta) -> str:
     return '%ds' % (seconds,)
 
 
+class IndentDumper(yaml.Dumper):
+  """A yaml dumper that indent lists."""
```
dsmilkov, line 287: will this dumper still write [a, b, c] for flat lists?

lilac/project.py
```diff
@@ -1,11 +1,104 @@
...
+                                                 dataset_config.name)
+    if existing_dataset_config is not None:
+      raise ValueError(f'{dataset_config} has already been added.')
+
+    config.datasets.append(dataset_config)
+    _write_project_config(data_path(), config)
+
+
+def delete_project_dataset_config(namespace: str, dataset_name: str) -> None:
+  """Add a dataset signal to the project config."""
```
dsmilkov, line 36: update pydoc here and all the methods below

dsmilkov (approved):
amazing work!

lilac/project.py
nsthorat: Done

lilac/utils.py
nsthorat: It will not -- there's no way to do both.

The older dumper ended up being completely illegible with JSON, and it's also apparently not standard formatting: https://reorx.com/blog/python-yaml-tips/

The old way wont work well with standard yml formatters (e.g. in vscode) and it's so hard to read the JSON.

The only time it's slightly less readable is for repeated paths, which I don't think is worth making the yaml file less readable overall.

Compare old:

image

To new:
image

lilac/data/dataset.py
nsthorat: Done

Issue #630, opened 2023-09-01
Tasks are broken in latest pip package


dsmilkov:
To repro:

`pip install lilac[all]` in a fresh virtual env. Then `lilac start my_project`, and open the browser. See the console output:

```
  File "/Users/dsmilkov/.pyenv/versions/3.11.4/envs/testpy3/lib/python3.11/site-packages/lilac/server.py", line 138, in _load
    task_manager=get_task_manager())
                 ^^^^^^^^^^^^^^^^^^
  File "/Users/dsmilkov/.pyenv/versions/3.11.4/envs/testpy3/lib/python3.11/site-packages/lilac/tasks.py", line 226, in get_task_manager
    return TaskManager()
           ^^^^^^^^^^^^^
  File "/Users/dsmilkov/.pyenv/versions/3.11.4/envs/testpy3/lib/python3.11/site-packages/lilac/tasks.py", line 106, in __init__
    self._dask_client = dask_client or Client(
                                       ^^^^^^^
  File "/Users/dsmilkov/.pyenv/versions/3.11.4/envs/testpy3/lib/python3.11/site-packages/distributed/client.py", line 960, in __init__
    self._loop_runner = LoopRunner(loop=loop, asynchronous=asynchronous)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dsmilkov/.pyenv/versions/3.11.4/envs/testpy3/lib/python3.11/site-packages/distributed/utils.py", line 519, in __init__
    raise RuntimeError(
RuntimeError: Constructing LoopRunner(asynchronous=True) without a running loop is not supported
```

nsthorat:
Hm, I'm unable to repro. Do you see this from poetry?

`poetry run lilac start my_project`?

dsmilkov:
Nope, I made a fresh virtual env and did pip install from pypi, started the server, and opened the browser, refreshed the home page and saw the console error.

TheodoreGalanos:
I have the same issue with `pip install lilac[gte]`, I'm on python 3.11

PR #635, opened 2023-09-01
Fix issue with task manager throwing an async error.

nsthorat:
This happens with the fastapi event loop is not yet ready. So we just throw an error and retry.

Fixes https://github.com/lilacai/lilac/issues/630

### DIFF
--- a/lilac/server.py
+++ b/lilac/server.py
@@ -127,7 +127,7 @@ def status() -> ServerStatus:
 
 
 @app.post('/load_config')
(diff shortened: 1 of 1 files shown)


dsmilkov:
I just saw this. Can you look at https://github.com/lilacai/lilac/pull/636 instead? I think it's a simpler solution and gets at the root cause, which is that we schedule load via a background task, and fastapi won't give it a loop unless you mark the task as async

PR #641, opened 2023-09-03
Add first version for Dataset Insights

dsmilkov:
https://github.com/lilacai/lilac/assets/2294279/e5ca25ec-8fe0-4e13-af51-180140eddb42

### DIFF
--- a/web/blueprint/src/lib/view_utils.ts
+++ b/web/blueprint/src/lib/view_utils.ts
@@ -5,7 +5,6 @@ import {
   childFields,
   deserializePath,
   getField,
-  getFieldsByDtype,
   isConceptSignal,
   isSignalField,
   pathIncludes,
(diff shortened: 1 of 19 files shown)


web/blueprint/src/lib/view_utils.ts
```diff
@@ -76,69 +76,15 @@ export const DTYPE_TO_ICON: Record<DataType, typeof CarbonIcon> = {
...
-  return visibleFields.some(f => pathIsEqual(f.path, field.path));
-}
-
-export function isItemVisible(item: LilacValueNode, visibleFields: LilacField[]): boolean {
-  const field = L.field(item);
-  if (field == null) return false;
-  return isFieldVisible(field, visibleFields);
-}
-
-export function getVisibleSchema(
```
nsthorat, line 109: I think this broke the metadata on the RHS -- now all fields are visible (any reason to delete this? it basically allows us not to have to force every child to check for visibility)

image

nsthorat (approved):
High-level UX stuff:

- Can we show a total # rows at the top of insights
- Can we show the average text length / some other media field stats at the top of the media field card?
- Bug: After clicking on open dataset, there's a bunch of history that's added so back button needs many clicks to get back (see my inline comment)
- Can you add `sticky` to the media field header? When there's a lot of ocncepts and I scroll, I get lost about what field it's on (and there's no other indication):
image

- Can we not show the version in the unexpanded version? it seems irrelevant to most people (expand is fine, or maybe hover tooltip or something)
image
- Maybe "Max. span score" => "Max score"

PR #644, opened 2023-09-05
Add expandable metadata

dsmilkov:

https://github.com/lilacai/lilac/assets/2294279/6141d20e-d772-4292-ab9d-24b9df98ee5e

image


web/blueprint/src/lib/stores/datasetViewStore.ts
```diff
@@ -23,7 +23,6 @@ export interface DatasetViewState {
   datasetName: string;
 
   // Explicit user-selected columns.
-  selectedColumns: {[path: string]: boolean};
   expandedColumns: {[path: string]: boolean};
```
nsthorat, line 26: can you put a comment here about what expandedColumns means now?

nsthorat (approved):
Amazing work, simplifying!

UX stuff:
Can we give some more space so we can read the column names here:
image

Can we add some horizontal space so the button focus doesnt overlap:
image

Same vertically:
image

When we open a signal root, can we expand all children so we dont have to click every span to see the scores:
image

I know this is my fault, but while you're here can we put a italic "n/a" when a field is empty:
image

Tiny regression here (colors dont appear on hover):
image

web/blueprint/src/lib/stores/datasetViewStore.ts
dsmilkov: done

dsmilkov:
all comments done, except the signal preview hover tooltip regression (it's from earlier)

PR #667, opened 2023-09-08
Add the python API for adding labels.

nsthorat:
New API on `Dataset` to add a label to a row or set of rows.
```py
dataset.add_label(
    label_name: str,
    label_value: str,
    searches: Optional[Sequence[Search]] = None,
    filters: Optional[Sequence[FilterLike]] = None) -> None:
```

When adding a label, labels are written to `{label_name}.label.jsonl` which contains json lines of the labels. This is not optimal, but we can switch out the implementation later.

This then acts as a regular feature, with `(label_name, 'label')` to access the actual label value. We also expose other metadata like `(label_metadata, 'created')` which tells you the timestamp when the label was created.

This also adds the [`freezegun`](https://github.com/spulec/freezegun) dev-dependency for freezing time in python tests. `freezegun` is a very popular library since python doesn't allow mocking datetime.now().


lilac/data/dataset_duckdb.py
```diff
@@ -200,25 +209,38 @@ def _recompute_joint_table(self, latest_mtime_micro_sec: int) -> DatasetManifest
...
+          with open_file(os.path.join(root, file)) as f:
+            signal_manifest = SignalManifest.parse_raw(f.read())
+          self._signal_manifests.append(signal_manifest)
+          signal_files = [os.path.join(root, f) for f in signal_manifest.files]
+          if signal_files:
+            self._create_view(signal_manifest.parquet_id, signal_files, type='parquet')
+        elif file.endswith(LABELS_FILENAME_SUFFIX):
+          label_name = file[0:-len(LABELS_FILENAME_SUFFIX)]
+          self._create_view(label_name, [os.path.join(root, file)], type='json')
+          # This mirrors the structure in DuckDBDatasetLabel.
```
dsmilkov, line 231: should we move `DatasetLabel` to _duckdb.py and rename to `DuckDBDatasetLabel` as mentioned in the comment?

lilac/data/dataset_duckdb.py
```diff
@@ -1153,6 +1182,37 @@ def select_rows_schema(self,
...
+
+    # TODO(nsthorat): Use duckdb to write the labels for better performance.
+    for row_chunk in chunks(rows, size=10_000):
+      labels = [{
+        ROWID: row[ROWID],
+        name: DatasetLabel(label=label, created=created).dict()
+      } for row in row_chunk]
+
+      # Write the chunk to the jsonl file, row by row.
+      with open(labels_filepath, 'a') as f:
```
dsmilkov, line 1212: I think we can speedup writing by orders of magnitude if we open it once outside the for loop? To avoid any thread problems, we can add a lock

dsmilkov (approved):
amazing, good start!

lilac/data/dataset_duckdb.py
nsthorat: DatasetLabel is going to be part of public REST API

lilac/data/dataset_duckdb.py
nsthorat: Done

PR #670, opened 2023-09-13
Add UI for viewing labels.

nsthorat:
NOTE: No UI has been added for *creating* labels, these labels were created from python. We are just rendering them now.

Demo: https://lilacai-nikhil-staging.hf.space/datasets#local/imdb&schemaCollapsed=false

Labels were added with:

```python
imdb.add_labels('readability', 'high', filters=[('text.text_statistics.readability', 'greater', 200)])
```

Rename `add_label` => `add_labels`

image

image

image


lilac/data/dataset.py
```diff
@@ -6,7 +6,10 @@
 import pathlib
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
-from typing import Any, Iterator, Literal, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Any, Iterator, Literal, Optional, Sequence, Union
+
+if TYPE_CHECKING:
```
dsmilkov, line 11: do you need this since it's a no-op?

dsmilkov:
This is a good start! I think we should soon render label in a special way (as tags on top of docs) just to simplify the UI, but reusing the generalized metadata-viewer is good for now

dsmilkov (approved):

lilac/data/dataset.py
nsthorat: Ha, I copy pasted a block and then the import got removed, so the auto-saver made it "pass"

PR #671, opened 2023-09-13
Fix bug with the selectRows cache not being cleared when labeling concepts in HF.

nsthorat:
This only happens in HF because we have a production-deployment that changes the names of functions!

Throwback to closure compilation issues ;)


web/blueprint/src/lib/queries/datasetQueries.ts
```diff
@@ -116,22 +116,24 @@ DATASETS_TAG);
...
+      datasetName: string,
+      selectRowsOptions: SelectRowsOptions
+    ) {
+      const res = await DatasetsService.selectRows(namespace, datasetName, selectRowsOptions);
+      return {
+        rows: schema == null ? res.rows : res.rows.map(row => deserializeRow(row, schema)),
+        total_num_rows: res.total_num_rows
+      };
+    },
+    [DATASETS_TAG, 'selectRows', namespace, datasetName]
```
dsmilkov, line 135: maybe we don't need to specify namespace and datasetnmae in tags, since createAPIQUery will add all args at the end ?  I guess it would make a difference if we ever do invalidateQuery('selectrows', namespace, datasetName)

dsmilkov (approved):
nice find!

web/blueprint/src/lib/queries/datasetQueries.ts
nsthorat: Yeah, I did it this way to mirror what we do for infiniteSelectRows (and in general we should probably try to invalidate exactly what needs to be invalidate, generally we just invalidate all datasets).

PR #678, opened 2023-09-15
Use sqlite instead of jsonl for labels.

nsthorat:


lilac/data/dataset_duckdb.py
```diff
@@ -266,7 +277,12 @@ def _recompute_joint_table(self, latest_mtime_micro_sec: int) -> DatasetManifest
     label_column_selects = []
     for label_name in self._label_schemas.keys():
       col_name = _escape_col_name(label_name)
-      label_column_selects.append(f'{col_name}.{col_name} AS {col_name}')
+      label_column_selects.append(f"""
```
dsmilkov, line 280: add a comment for why we need this special case, is it because labels are sparse?

lilac/data/dataset_duckdb.py
```diff
@@ -1213,18 +1229,25 @@ def add_labels(self,
...
+          {ROWID} VARCHAR NOT NULL PRIMARY KEY,
+          label VARCHAR NOT NULL,
+          created DATETIME)
+      """)
+
+      for row in rows:
+        sqlite_cur.execute(
+          f"""
+            INSERT INTO {name} VALUES (?, ?, ?)
+            ON CONFLICT({ROWID}) DO UPDATE SET label=excluded.label;
```
dsmilkov, line 1248: add a comment on what this does

dsmilkov (approved):

lilac/data/dataset_duckdb.py
nsthorat: Done

lilac/data/dataset_duckdb.py
nsthorat: Done

PR #688, opened 2023-09-18
Separate metadata requests into a single request for row_ids, and multiple for metadata.

nsthorat:
This allows us to invalidate just a single row for labeling a single row.

https://lilacai-nikhil-staging.hf.space/datasets#local/glue

https://github.com/lilacai/lilac/assets/1100749/32174191-65e5-4c0e-af26-d58040c00629

### DIFF
--- a/web/blueprint/src/lib/components/datasetView/AddLabel.svelte
+++ b/web/blueprint/src/lib/components/datasetView/AddLabel.svelte
@@ -45,7 +45,7 @@
       .map((l, i) => ({id: `label_${i}`, text: l})) || [];
   $: labelItems = [...(comboBoxText != '' ? [newLabelItem] : []), ...missingLabelItems];
 
-  const addLabels = addLabelsMutation();
+  const addLabels = $datasetStore.schema != null ? addLabelsMutation($datasetStore.schema) : null;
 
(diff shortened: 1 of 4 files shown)


web/blueprint/src/lib/components/datasetView/AddLabel.svelte
```diff
@@ -45,7 +45,7 @@
       .map((l, i) => ({id: `label_${i}`, text: l})) || [];
   $: labelItems = [...(comboBoxText != '' ? [newLabelItem] : []), ...missingLabelItems];
 
-  const addLabels = addLabelsMutation();
+  $: addLabels = $datasetStore.schema != null ? addLabelsMutation($datasetStore.schema) : null;
```
dsmilkov, line 48: curious when schema would be null? if you try to click label too quickly? or does it have to do with the invalidation of schema?

dsmilkov (approved):

web/blueprint/src/lib/components/datasetView/AddLabel.svelte
nsthorat: it's null just because it's async -- the datasetstore starts off with a null schema, when the request comes back then only we have the schema

PR #689, opened 2023-09-19
Fix upload scripts so we can push to a staging directory without uploading data.

nsthorat:
This was forked from the public space, and then pushed with:

```
poetry run python -m scripts.deploy_demo \
--use_pip=false \
--skip_sync \
--skip_load \
--skip_data_upload \
--hf_space=lilacai/lilac-staging
```

https://huggingface.co/spaces/lilacai/lilac-staging

### DIFF
--- a/scripts/deploy_demo.py
+++ b/scripts/deploy_demo.py
@@ -25,7 +25,7 @@ poetry run python -m scripts.deploy_demo \
   --use_pip=false \
   --skip_sync \
   --skip_load \
-  --skip_upload \
+  --skip_data_upload \
   --use_pip=false \
   --hf_space=lilacai/lilac-staging
 """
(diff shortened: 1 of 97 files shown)


scripts/deploy_demo.py
```diff
@@ -69,9 +85,13 @@
...
   is_flag=True,
   default=False)
+@click.option(
+  '--use_pip',
+  help='When true, uses the public pip package. When false, builds and uses a local wheel.',
+  default=True)
 def deploy_demo(config: str, hf_space: str, project_dir: str, overwrite: bool, skip_sync: bool,
-                skip_load: bool, skip_build: bool, skip_deploy: bool,
-                make_datasets_public: bool) -> None:
+                skip_load: bool, skip_build: bool, skip_upload: bool, skip_deploy: bool,
```
dsmilkov, line 93: skip_data_upload

dsmilkov (approved):
awesome!!

scripts/deploy_demo.py
nsthorat: done

PR #691, opened 2023-09-19
Add better error messaging when inferring schema

dsmilkov:
Also fix a bug where serializing `Source` didn't serialize any specific fields due to Pydantic V2 migration.

### DIFF
--- a/lilac/utils.py
+++ b/lilac/utils.py
@@ -240,7 +240,7 @@ def log(*args: Any) -> None:
     return
 
   print(*args)
-  logging.info([' '.join(args)])
+  logging.info([' '.join(str(args))])
 
 
 class DebugTimer:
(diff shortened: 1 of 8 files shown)


lilac/utils.py
```diff
@@ -234,13 +234,13 @@ def is_primitive(obj: object) -> bool:
...
 
-def log(*args: str) -> None:
+def log(*args: Any) -> None:
   """Print and logs a message so it shows up in the logs on cloud."""
   if env('DISABLE_LOGS'):
     return
 
   print(*args)
-  logging.info([' '.join(args)])
+  logging.info([' '.join(str(args))])
```
nsthorat, line 243: str(args) => str, so the join has a non-array here, intended?

nsthorat (approved):

lilac/utils.py
dsmilkov: nice catch, done

PR #698, opened 2023-09-21
Allow `add_labels` and `remove_labels` without selection

dsmilkov:


lilac/data/dataset.py
```diff
@@ -258,19 +259,32 @@ class Filter(BaseModel):
 SearchValue = StrictStr
 
 
+def _fix_const_in_schema(prop_name: str, value: str) -> Callable[[dict[str, Any]], None]:
```
nsthorat, line 262: maybe something more like "_replace_prop_with_enum" instead of "fix"

nsthorat (approved):

PR #701, opened 2023-09-21
Make '.' the default project.

nsthorat:
image

### DIFF
--- a/lilac/env.py
+++ b/lilac/env.py
@@ -7,6 +7,9 @@ from dotenv import load_dotenv
 from pydantic import BaseModel
 from pydantic import Field as PydanticField
 
+# The project directory defaults to the current working directory.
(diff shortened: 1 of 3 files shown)


lilac/env.py
```diff
@@ -108,9 +111,9 @@ def env(key: str, default: Optional[Any] = None) -> Any:
 
 def get_project_dir() -> str:
   """Return the base path for data."""
-  project_dir = env('LILAC_PROJECT_DIR', None)
```
dsmilkov, line 111: Should this default to none to keep reading the old env var?

dsmilkov (approved):

lilac/env.py
nsthorat: yep

Issue #660, opened 2023-09-06
Memory leak/pressure when generating embedding for large dataset


chilang:
Hi,
I'm getting what seems to be memory leak warnings when completing the embedding generation for multi-million items datasets. 
The logs look like below and the memory seems to eventually get released. 

However, for some larger datasets 4m+ items like OpenOrca, it can use up all available RAM (e.g. 72GB) at the end of the embedding generation but before writing down the embedding and crash the main web serving lilac process. When this happens, the generated embedding is lost - UI shows as if there is no embedding.

```
[local/arxiv] Compute signal "gte-small" on "abstract": 100%|█████████████████████████████████████████████████████████████████████████████████████| 2314976/2314976 [1:33:40<00:00, 411.87it/s]
Computing signal "gte-small" on local/arxiv:('abstract',) took 5623.462s.
2023-09-06 12:19:04,702 - distributed.worker.memory - WARNING - Unmanaged memory use is high. This may indicate a memory leak or the memory may not be released to the OS; see https://distributed.dask.org/en/latest/worker-memory.html#memory-not-released-back-to-the-os for more information. -- Unmanaged memory: 46.01 GiB -- Worker memory limit: 65.73 GiB
2023-09-06 12:19:14,703 - distributed.worker.memory - WARNING - Unmanaged memory use is high. This may indicate a memory leak or the memory may not be released to the OS; see https://distributed.dask.org/en/latest/worker-memory.html#memory-not-released-back-to-the-os for more information. -- Unmanaged memory: 46.12 GiB -- Worker memory limit: 65.73 GiB
2023-09-06 12:19:24,703 - distributed.worker.memory - WARNING - Unmanaged memory use is high. This may indicate a memory leak or the memory may not be released to the OS; see https://distributed.dask.org/en/latest/worker-memory.html#memory-not-released-back-to-the-os for more information. -- Unmanaged memory: 46.23 GiB -- Worker memory limit: 65.73 GiB
2023-09-06 12:19:34,803 - distributed.worker.memory - WARNING - Unmanaged memory use is high. This may indicate a memory leak or the memory may not be released to the OS; see https://distributed.dask.org/en/latest/worker-memory.html#memory-not-released-back-to-the-os for more information. -- Unmanaged memory: 46.34 GiB -- Worker memory limit: 65.73 GiB
[...]

PR #706, opened 2023-09-21
Use iterables for compute_signal and compute_embedding.

nsthorat:
High-level:
- compute_signal() and compute_embedding() now directly call duckdb to get source values. We use `fetch_df_chunk` to iteratively get chunks of the response of the query.
- Downstream, everything is now iterable, which means all of the batch & dataset utils are now generators.

This is the first step towards https://github.com/lilacai/lilac/issues/660

A follow up will actually write embeddings iteratively.

Regression: now we do not support running signals over span fields. This isn't really supported anywhere, but easy to fix if we decide we want it later.

https://lilacai-nikhil-staging.hf.space/

### DIFF
--- a/lilac.yml
+++ b/lilac.yml
@@ -1,0 +1,4 @@
+# Lilac project config.
+# See https://lilacml.com/api_reference/index.html#lilac.Config for details.
+
+{}
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -361,13 +361,12 @@ class DatasetDuckDB(Dataset):
       self._vector_indices[index_key] = vector_index
       return vector_index
 
-  def _select_iterable_values(self, path: PathTuple) -> Iterable[tuple[str, RichData]]:
+  def _select_iterable_values(self, path: PathTuple,
+                              data_schema: Schema) -> Iterable[tuple[str, RichData]]:
     # Fetch the data from DuckDB.
     con = self.con.cursor()
 
-    manifest = self.manifest()
-
-    duckdb_path = self._leaf_path_to_duckdb_path(path, manifest.data_schema)
+    duckdb_path = self._leaf_path_to_duckdb_path(path, data_schema)
 
     sql = _select_sql(duckdb_path, flatten=False, unnest=False, empty=False)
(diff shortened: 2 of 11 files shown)


lilac.yml
dsmilkov, line 1: do we need this in source code repo?

scripts/test_py.sh
```diff
@@ -14,8 +14,6 @@ if [ "$CI" ]; then
   PYTEST_MARKS="(not largedownload)"
 fi
 
-# Disables log() statements.
-export DISABLE_LOGS=True
```
dsmilkov, line 18: intentional? probs a good idea right?

lilac/data/dataset_duckdb.py
```diff
@@ -429,37 +506,58 @@ def compute_embedding(self,
...
     gc.collect()
 
+    signal_manifest_filepath = os.path.join(output_dir, SIGNAL_MANIFEST_FILENAME)
+    # If the signal manifest already exists, delete it as it will be rewritten after the new signal
+    # outputs are run.
+    if os.path.exists(signal_manifest_filepath):
+      os.remove(signal_manifest_filepath)
+      # Call manifest() to recreate all the views, otherwise this could be stale and point to a non
+      # existent file.
+      self.manifest()
```
dsmilkov, line 552: why do you need a new manifest here? does seem like it's used below in this current function?

dsmilkov (approved):
amazing work!

lilac.yml
nsthorat: nice, mistake.

lilac/data/dataset_duckdb.py
nsthorat: See the comment above but basically this can happen if you run a signal twice and make a request to the server.

What happens:
1) first signal runs, completes, all is well
2) Run it again. We just deleted the signal manifest so that nobody else can query while we're building a new file. If you don't call manifest() here, we still have an old view, and thus tries to query an empty parquet file and fails.

scripts/test_py.sh
nsthorat: Intentional, and a good idea.

PR #711, opened 2023-09-24
Add clustering signal

dsmilkov:
Also improve the vector store fetching by ~6x by batching all the vector keys in a single `get` call. Clustering ~50,000 chunks now takes 29sec on my laptop. 

https://github.com/lilacai/lilac/assets/2294279/2f4641df-b9fc-40d5-a5b5-8b6b8cac815e


lilac/signals/clustering.py
```diff
@@ -0,0 +1,61 @@
...
+
+    span_sizes: list[int] = []
+    all_vectors: list[np.ndarray] = []
+    with DebugTimer('getting vectors'):
+      for vectors in span_vectors:
+        span_sizes.append(len(vectors))
+        for vector in vectors:
+          all_vectors.append(vector['vector'])
+
+    dbscan = DBSCAN(eps=0.05, min_samples=MIN_SAMPLES, metric='cosine', n_jobs=-1)
```
nsthorat: FYI, there is a closely related algorithm called OPTICS which is used for much larger datasets without having to use an entirely in-memory embedding matrix, however it requires the nightly build of sklearn:
https://scikit-learn.org/dev/modules/generated/sklearn.cluster.OPTICS.html

Maybe we should put this as a comment somewhere to try it later

lilac/signals/clustering.py
```diff
@@ -0,0 +1,62 @@
...
+from ..schema import Field, Item, PathKey, RichData, SignalInputType, SpanVector, field
+from ..signal import VectorSignal
+
+CLUSTER_IDS = 'cluster_ids'
+MIN_SAMPLES = 5
+DBSCAN_EPS = 0.05
+
+
+class ClusteringSignal(VectorSignal):
+  """Find clusters of documents in a dataset."""
```
nsthorat, line 21: maybe specify this uses embeddings and dbscan

lilac/signals/clustering.py
```diff
@@ -0,0 +1,62 @@
...
+from lilac.embeddings.vector_store import VectorDBIndex
+from lilac.utils import DebugTimer
+
+from ..embeddings.embedding import get_embed_fn
+from ..schema import Field, Item, PathKey, RichData, SignalInputType, SpanVector, field
+from ..signal import VectorSignal
+
+CLUSTER_IDS = 'cluster_ids'
+MIN_SAMPLES = 5
+DBSCAN_EPS = 0.05
```
nsthorat, line 17: thoughts on this being an argument to the signal?

pros and cons: nice to configure, but exposes impl details.

and then that begs the question should we hide the impl details?

nsthorat (approved):

lilac/signals/clustering.py
dsmilkov: done

lilac/signals/clustering.py
dsmilkov: done

PR #715, opened 2023-09-25
Improve perf of server by removing UUID sort

dsmilkov:
- Add sort by UUID in the UI
- Return number of labels via the REST endpoint
- Fix SQLite escape error


nsthorat (approved):
amazing

PR #717, opened 2023-09-25
Change the bootup load to be during the new FastAPI lifecycle API.

nsthorat:
- `LILAC_LOAD_ON_START_SERVER` controls whether loading the lilac.yml gets loaded on start.
- Loading now happens in the lifecycle event, which means we do this gunicorn or start_server
- The load() call itself is now is a task, and it schedules more tasks. This puts it in the top-right tasks menu.

We can now just edit a lilac.yml in a HF space with the bit set and it will load the dataset!

Demo (this is a manually edited config that loads): https://huggingface.co/spaces/nsthorat-lilac/nikhil_staging2


lilac/server.py
```diff
@@ -65,10 +65,41 @@ def custom_generate_unique_id(route: APIRoute) -> str:
...
+    load_task_id=load_task_id)
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
+  """Context manager for the lifespan of the application."""
+  if env('LILAC_LOAD_ON_START_SERVER', False):
+    task_manager = get_task_manager()
+    loop = asyncio.get_running_loop()
+    loop.run_in_executor(None, _add_load_task, task_manager)
```
dsmilkov, line 90: maybe call task_manager.execute directly since it's non blocking

dsmilkov (approved):

lilac/server.py
nsthorat: Done

PR #726, opened 2023-09-29
Improve the schema view to be less overwhelming.

nsthorat:
Use the signal titles & concept names instead of the paths for signal roots.

Definitely alot more we can do, but step in the right direction.

After:
image

Hover chips show the embeddings & signal metadata:
image

Before:

image


web/blueprint/src/lib/components/schemaView/SchemaField.svelte
```diff
@@ -49,8 +51,34 @@
...
-  $: fieldName = isRepeatedField ? path.at(-2) : path.at(-1);
+  $: fieldTitle = isRepeatedField ? path.at(-2) : path.at(-1);
+  let fieldHoverDetails: string | null = null;
+  // If the field is a signal root, use the signal name to define the title.
+  $: {
+    if (field.signal && $signals.data != null) {
+      if (field.signal.signal_name === 'concept_score') {
+        const conceptSignal = field.signal as ConceptSignal;
+        fieldTitle = `${conceptSignal.concept_name}`;
+        fieldHoverDetails = `Concept '${conceptSignal.concept_name}' arguments:\n\nembedding: '${conceptSignal.embedding}'`;
```
dsmilkov, line 65: should we show the version number if we compute v17 and v18 of these.... Alternatively, and I think simpler to do is to always output "namespace/concept" as the column, without the version number, so re-running the signal overwrites the column. We can still have an "outdated" badge by comparing the written version in `signal_manifest.signal.version`  with the latest concept version

dsmilkov (approved):

web/blueprint/src/lib/components/schemaView/SchemaField.svelte
nsthorat: Did the second approach -- now we emit: namespace/name/embedding

version is now in the signal, and shown in the hover

dsmilkov: amazing

PR #742, opened 2023-10-04
Extract the dropdown+pill into a `DropPill` component

dsmilkov:


web/blueprint/src/lib/components/common/DropPill.svelte
```diff
@@ -0,0 +1,84 @@
+<!-- A combination of a dropdown and filter-like pill -->
```
nsthorat, line 1: maybe call this DropdownPill? Might be confusing for new folks later

nsthorat (approved):
Waiting for the demo...

nsthorat (commented):
What do you think about using the Folder icon instead for Group, since it feels like you're browsing folders (the other icon is kind of confusing)

image

PR #744, opened 2023-10-04
Make llama-index an optional dependency.

nsthorat:


pyproject.toml
```diff
@@ -94,6 +93,7 @@ all = [
   "langdetect",
   "spacy",
   "langsmith",
+  "github",
```
dsmilkov, line 96: i'm not sure you can link to aliases here (e.g. github). you might need to link directly to the pip package

dsmilkov (approved):

pyproject.toml
nsthorat: Done

PR #748, opened 2023-10-05
Make labels toggle-able 

dsmilkov:

Also:
- do css fixes
- remove the sqlite db when all the instances of a label are gone.

### DIFF
--- a/web/blueprint/src/lib/components/datasetView/RowItem.svelte
+++ b/web/blueprint/src/lib/components/datasetView/RowItem.svelte
@@ -52,8 +52,7 @@
       : null;
   $: row = $rowQuery?.data;
   $: rowLabels = row != null ? getRowLabels(row) : [];
-  $: disableLabels =
-    !canEditLabels || $removeLabels?.isLoading || $datasetStore.selectRowsSchema?.isFetching;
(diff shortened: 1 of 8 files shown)


web/blueprint/src/lib/components/datasetView/RowItem.svelte
```diff
@@ -38,14 +50,51 @@
...
-  $: row = $rowQuery?.data != null ? $rowQuery.data : null;
+  $: row = $rowQuery?.data;
   $: rowLabels = row != null ? getRowLabels(row) : [];
+  $: disableLabels = !canEditLabels;
 
-  const removeLabel = (label: string) => {
+  $: schemaLabels = $datasetStore.schema && getSchemaLabels($datasetStore.schema);
+  $: addLabels = $datasetStore.schema != null ? addLabelsMutation($datasetStore.schema) : null;
+
+  $: isStale = $rowQuery?.isStale;
```
nsthorat, line 60: might be worth a comment here about why we derive this, i can see myself being like "ill just make this code shorter by using $rowQuery.isStale directly"

nsthorat (approved):

PR #749, opened 2023-10-08
Add HDBScan with UMAP

nsthorat:
Uses UMAP to reduce the dimensionality before running HDBScan, greatly improving the performance of clustering.

### DIFF
--- a/lilac/signals/cluster_hdbscan.py
+++ b/lilac/signals/cluster_hdbscan.py
@@ -22,7 +22,7 @@ DBSCAN_EPS = 0.05
 class ClusterHDBScan(VectorSignal):
   """Find clusters of documents in a dataset using pre-computed embeddings and DBSCAN."""
   name: ClassVar[str] = 'cluster_hdbscan'
-  display_name: ClassVar[str] = 'Cluster with HDDBSCAN'
+  display_name: ClassVar[str] = 'Cluster with HDBSCAN'
   input_type: ClassVar[SignalInputType] = SignalInputType.TEXT
 
   min_cluster_size: int = PyField(
--- a/lilac/signals/cluster_hdbscan.py
+++ b/lilac/signals/cluster_hdbscan.py
@@ -3,8 +3,6 @@ from typing import ClassVar, Iterable, Optional
 
 import numpy as np
 import umap
-
-# from hdbscan import HDBSCAN
 from pydantic import Field as PyField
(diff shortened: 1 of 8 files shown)


lilac/signals/cluster_hdbscan.py
```diff
@@ -0,0 +1,81 @@
...
+          all_vectors.append(vector['vector'])
+
+    # Use UMAP to reduce the dimensionality before hdbscan to speed up clustering.
+    # For details on hyperparameters, see:
+    # https://umap-learn.readthedocs.io/en/latest/clustering.html
+    reducer = umap.UMAP(n_components=10, n_neighbors=30, min_dist=0.0)
+    all_vectors = reducer.fit_transform(all_vectors)
+
+    hdbscan = HDBSCAN(min_cluster_size=self.min_cluster_size, n_jobs=-1)
+    hdbscan.fit(np.array(all_vectors))
```
dsmilkov, line 68: nit: no need to wrap in np.array() since all_vectors comes from umap

lilac/signals/cluster_hdbscan.py
```diff
@@ -0,0 +1,81 @@
...
+    with DebugTimer('DBSCAN: Reading from vector store'):
+      for vectors in span_vectors:
+        all_spans.append([vector['span'] for vector in vectors])
+        for vector in vectors:
+          all_vectors.append(vector['vector'])
+
+    # Use UMAP to reduce the dimensionality before hdbscan to speed up clustering.
+    # For details on hyperparameters, see:
+    # https://umap-learn.readthedocs.io/en/latest/clustering.html
+    reducer = umap.UMAP(n_components=10, n_neighbors=30, min_dist=0.0)
```
dsmilkov, line 64: extract n_components into a constant since it's the param that balances speed/accuracy

lilac/signals/cluster_hdbscan.py
```diff
@@ -0,0 +1,81 @@
+"""Compute clusters for a dataset."""
+from typing import ClassVar, Iterable, Optional
+
+import numpy as np
+import umap
```
dsmilkov, line 5: since we made umap-learn optional , import it in a try/catch inside compute().

Alternatively, and maybe better is to make umap-learn required since we will probably use it in "Cluster View" when we add 2d plots

dsmilkov (approved):

Issue #737, opened 2023-10-04
Simple docker command in readme / getting started


josephrocca:
This project is really neat! I'm using it to help build a toxicity classifier for a comments moderation system.

Question: Can you add a simple docker command to the readme or quickstart/installation guide? If I click the three vertical dots on the Hugging Face space and click "Run with Docker" (in the same menu as the "Duplicate this Space" button) then it gives me this command:
```
docker run -it -p 7860:7860 --platform=linux/amd64 \
	-e HF_ACCESS_TOKEN="YOUR_VALUE_HERE" \
	-e GOOGLE_CLIENT_SECRET="YOUR_VALUE_HERE" \
	-e LILAC_OAUTH_SECRET_KEY="YOUR_VALUE_HERE" \
	-e DUCKDB_USE_VIEWS="1" \
	-e HF_HOME="/data/.huggingface" \
	-e _SEE_LILAC_DOCS_="https://lilacml.com/huggingface/huggingface_spaces.html" \
	-e LILAC_PROJECT_DIR="/data" \
	-e TRANSFORMERS_CACHE="/data/.cache" \
	-e XDG_CACHE_HOME="/data/.cache" \
	-e GOOGLE_ANALYTICS_ENABLED="true" \
	-e LILAC_AUTH_ENABLED="true" \
	-e GOOGLE_CLIENT_ID="279475920249-i8llm8vbos1vj5m1qocir8narb3r0enu.apps.googleusercontent.com" \
	registry.hf.space/lilacai-lilac:latest
```
Which I trimmed down to this since I don't need any auth/access-token stuff:
```
docker run -it -p 7860:7860 --platform=linux/amd64 \
	-e DUCKDB_USE_VIEWS="1" \
	-e HF_HOME="/data/.huggingface" \
	-e _SEE_LILAC_DOCS_="https://lilacml.com/huggingface/huggingface_spaces.html" \
	-e LILAC_PROJECT_DIR="/data" \
	-e TRANSFORMERS_CACHE="/data/.cache" \
	-e XDG_CACHE_HOME="/data/.cache" \
	-e LILAC_AUTH_ENABLED="false" \
	registry.hf.space/lilacai-lilac:latest
```
But for some reason the port was wrong - I had to switch it to `5432`. And I also had to add a `--volume` mount.

So it wasn't too hard to get working, but it would be great it the readme could have a simple, trimmed-down docker command to make it easier for people like me to quickly try out Lilac. Thanks!

[...]

PR #750, opened 2023-10-09
Add lilac docker image and use a lighter 5x image for HF

dsmilkov:
CleanShot 2023-10-09 at 13 54 23@2x

Fixes https://github.com/lilacai/lilac/issues/737

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update --fix-missing && apt-get install -y --fix-missing build-essen
 # Set the working directory in the container.
 WORKDIR /app
 
+# To reduce the image size, replace lilac[all] with lilac[...] optional deps.
 RUN python -m pip install lilac[all]
 
 # Install from the local wheel inside ./dist. This will be a no-op if the wheel is not found.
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ENV PYTHONUNBUFFERED True
+
+# Adds GCC and other build tools so we can compile hnswlib and other native/C++ deps.
+RUN apt-get update --fix-missing && apt-get install -y --fix-missing build-essential && \
+  rm -rf /var/lib/apt/lists/*
+
+# Set the working directory in the container.
+WORKDIR /app
+
+RUN python -m pip install lilac[all]
```
nsthorat, line 13: maybe comment here about optional deps and reducing size

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Install from the local wheel inside ./dist. This will be a no-op if the wheel is not found.
+COPY --chown=user /dis[t] ./dist/
+RUN python -m pip install --find-links=dist --upgrade lilac[all]
+
+COPY LICENSE .
+
+EXPOSE 8000
```
nsthorat, line 21: arg here? https://docs.docker.com/engine/reference/builder/#arg

nsthorat (approved):

Dockerfile
nsthorat: nevermind, you clarified that this is just for exposing & documentation

Dockerfile
dsmilkov: done

PR #755, opened 2023-10-11
Add CLI prompts for token in deploy script

nsthorat:
Instead of forcing the user to run the CLI, get the token, and put it in .env, we now prompt the user to login and ask them to copy paste from the console, which is simpler to deploy (since firebase tokens keep expiring)

image


dsmilkov (approved):
nice!

PR #760, opened 2023-10-12
Tiny UI fixes

dsmilkov:
Fix the error where NumberInput values didn't propagate to the server. Also remove the no-longer used version field in concept_score signal

### DIFF
--- a/lilac/signals/concept_scorer.py
+++ b/lilac/signals/concept_scorer.py
@@ -23,8 +23,6 @@ class ConceptSignal(VectorSignal):
 
   namespace: str
   concept_name: str
-  # This will get filled out during setup.
-  version: Optional[int] = None
 
   # The draft version of the concept to use. If not provided, the latest version is used.
   draft: str = DRAFT_MAIN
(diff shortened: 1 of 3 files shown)


lilac/signals/concept_scorer.py
```diff
@@ -69,11 +67,6 @@ def _compute_span_vector_batch(self, span_vectors: Iterable[SpanVector],
     scores = concept_model.score_embeddings(self.draft, np.array(vectors)).tolist()
     return [lilac_span(start, end, {'score': score}) for score, (start, end) in zip(scores, spans)]
 
-  @override
-  def setup(self) -> None:
```
nsthorat, line 73: revert -- this is on purpose, it shows up in the UI on hover so you know the version

nsthorat (approved):

lilac/signals/concept_scorer.py
dsmilkov: done, i added a comment for context since python doesn't use it. 

The reason for the motivation was that you get this version field in Compute Concept modal:
CleanShot 2023-10-12 at 10 18 09@2x

PR #762, opened 2023-10-12
Batch row metadata requests.

nsthorat:
See official tanstack docs that point to this library: https://tanstack.com/query/v4/docs/react/community/batching-requests-using-bathshit

Demo: https://lilacai-nikhil-staging.hf.space/datasets#lilac/OpenOrca-100k

Also fix some of the taglines.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,7 +1,7 @@
 🌸 Lilac
 =================
 .. meta::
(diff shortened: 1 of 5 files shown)


nsthorat:
Spoke offline - this is because you added a new label which invalidates the manifest. Adding an existing label only invalidates a single row!

PR #764, opened 2023-10-12
Improve dataset load onboarding flow.

nsthorat:
When loading a dataset, show the field selection settings, improve the wording around it. Add a preview at the bottom.

Remove the datasetStore entirely. This is useless, and causes issues when using components from outside dataset pages.

https://github.com/lilacai/lilac/assets/1100749/2f771ddc-62ac-4131-95f3-5bb2b51a2f45


lilac/data/dataset.py
```diff
@@ -55,6 +55,8 @@
 SAMPLE_AVG_TEXT_LENGTH = 1000
 MAX_TEXT_LEN_DISTINCT_COUNT = 250
 
+DEFAULT_EMBEDDING = 'gte-small'
```
dsmilkov, line 58: should we do EMBEDDING_SORT_PRIORITIES[0] ?

lilac/data/dataset_config_test.py
```diff
@@ -95,7 +95,8 @@ def test_config_compute_signal(make_test_data: TestDataMaker) -> None:
     name='test_dataset',
     source=TestSource(),
     # 'text' is the longest path, so should be set as the default setting.
-    settings=DatasetSettings(ui=DatasetUISettings(media_paths=[('text',)])))
+    settings=DatasetSettings(
+      ui=DatasetUISettings(media_paths=[('text',)]), preferred_embedding=DEFAULT_EMBEDDING))
```
dsmilkov, line 99: instead of having to define it explicitly, should we default it to DEFAULT_EMB in the basemodel definition of DatasetUISetting

dsmilkov (approved):

lilac/data/dataset.py
nsthorat: Done

lilac/data/dataset_config_test.py
nsthorat: Done

Issue #761, opened 2023-10-12
Can't... do anything


enn-nafnlaus:
Just thought I'd give lilac a try for cleaning my datasets.  Unfortunately I hit a stumbling block right off the bat:

![image](https://github.com/lilacai/lilac/assets/116288799/f79d2899-c3cd-4643-96ea-5e3e885ee5af)

First I couldn't search by output, but I found how to enable that in "administration"You'll note that my dataset has "instruction" and "output". But when I find something I'm not happy with... there's nothing I can do with I can't edit it.  I can't delete it.  It's just... there, in the search results.  The only thing I seem to have an option to do is "Label all", but that fails

![image](https://github.com/lilacai/lilac/assets/116288799/3757f4de-1c9f-4c13-8948-9b81e6fb9b4f)

Lilac is running as the same user that created and owns the json files, there's no issues with them not being accessible.

ED: Looks like I can create individual labels, just not label all.  But what good is that?  If there's bad data, I need to edit or delete it.

nsthorat:
@enn-nafnlaus so sorry that this is happening.

Is this from a fork of the HuggingFace space, or are you running it locally?

Can you remove the LILAC_AUTH_ENABLED environment variable which is disabling all of these features (either locally or on HF)?

enn-nafnlaus:
Running locally.
LILAC_AUTH_ENABLED is not set in my environment

I restarted and was able to do label_all, but labeling is still the only feature I'm able to do. Which just created a new problem, as in my test I did label_all "Test", and all 40k rows got labeled "Test", but now I can't remove the labels except one at a time.

nsthorat:
Thanks for the feedback -- we'll add a batch delete and improve onboarding with media / metadata fields.

For now, you can delete all of those labels if into the project directory you can delete the labels sqlite file (e.g. my_label.labels.sqlite under project_dir/datasets/local/summarize) and refresh (label all is going to label the entire slice that's currently selected, with no filters it's the entire dataset).

You can also remove a subset of them from python with: [`dataset.remove_labels`](https://lilacml.com/api_reference/data.html#lilac.data.Dataset.remove_labels)

enn-nafnlaus:
Thanks.  :)

When I saw this project recommended on the Axolotl discord, my impression was that one could search, and then just edit or delete the results... but I guess for now it's "display and label only".

[...]

PR #768, opened 2023-10-16
Add `include_labels` and `exclude_labels` when exporting data via `dataset.to_*`

dsmilkov:
This enables the user to download a subset of the dataset where some labels are gone.

https://lilac-386213--staging-h3b3igsj.web.app/datasets/dataset_export.html#labels

Towards https://github.com/lilacai/lilac/issues/761

### DIFF
--- a/docs/datasets/dataset_export.md
+++ b/docs/datasets/dataset_export.md
@@ -22,47 +22,51 @@ Click "Download" to download the results as a JSON file from the browser.
 
 ## From Python
 
-In Python, we can call `select_rows` to download the JSON. Let's download `text` and
-`text.language_detection`:
+In Python, we can export to different formats using the [](#Dataset.to_pandas),
+[](#Dataset.to_json), [](#Dataset.to_parquet) and [](#Dataset.to_csv) methods.
+
+Let's export the `text` and `text.language_detection` to a pandas dataframe:
 
 ```python
 dataset = ll.get_dataset('local', 'imdb')
-
-# Select 'text' and 'text.language_detection'.
-# NOTE: you can select lang detection with 'text.lang_detection' or ('text', 'language_detection')
-results = dataset.select_rows(['text', ('text', 'lang_detection'), 'text'])
-
-# Print the first result.
-# NOTE: `results` is an iterable of dicts (not a list).
-print(next(iter(results)))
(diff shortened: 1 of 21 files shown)


docs/datasets/dataset_export.md
```diff
@@ -22,47 +22,51 @@ Click "Download" to download the results as a JSON file from the browser.
...
-                              limit=limit,
-                              offset=offset)
-  print(list(items))
-
-  offset += limit
+dataset.to_json('imdb.jsonl', jsonl=True, filters=[('text.lilac/toxicity/gte-small.score', 'less', 0.3)])
 ```
 
-Output (each row is a chunk of results):
+### Labels
```
nsthorat, line 59: maybe put this before filters

scripts/deploy_website.py
```diff
@@ -23,25 +20,15 @@
...
-@click.option(
-  '--firebase_token',
-  help='The firebase token for authentication. Generate this with `firebase login:ci`.',
-  type=str,
-  required=False)
-def main(staging: bool, firebase_token: Optional[str] = None) -> None:
+def main(staging: bool) -> None:
   """Generate a web client from the OpenAPI spec."""
-  firebase_token = firebase_token or env('FIREBASE_TOKEN')
-  if not firebase_token:
```
nsthorat, line 34: i wouldnt revert this -- this makes it much much simpler if your token is expired. the other option is quite annoying (getting token independently and then putting it in ENV)

nsthorat (approved):
nice work, lgtm

scripts/deploy_website.py
dsmilkov: Service account keys never expire, whereas tokens are 24hr only. Also it says --token is deprecated and will be removed in a future version and will be replaced with service account keys

nsthorat: Gotchya, can you document how you do use the service account?

docs/datasets/dataset_export.md
dsmilkov: done

scripts/deploy_website.py
dsmilkov: I reverted and always call `firebase auth --reauth` when you deploy -- works all the time, and re-logging is that bad.

Issue #770, opened 2023-10-19
Intelligent sampling for large datasets


jtalmi:
Lilac currently does not work well with large datasets living in cloud storage. We often need to inspect samples of multi-GB/TB, partitioned parquet datasets but we aren't able to load the entires dataset locally. Ideally this would mean sampling a few rows from each shard, but a simpler version would be taking a single shard. 

If Lilac adds better support for sampling, it could be used to share/inspect/evaluate large pre-training datasets

nsthorat:
We've definitely been focusing on the single-node solution but we'll start thinking about larger datasets soon.

How big is one of your shards and in what format? We can currently read just a single shard (e.g. if it's parquet) if you use the parquet source. We have sampling for the HuggingFace source, but we can also add a sampling mechanism for arbitrary data.

jtalmi:
We try to keep shards small...maybe <200 MB. The main issue with taking a single shard is skewness, if the dataset isn't properly shuffled. An ideal solution would sample across all shards of a parquet dataset, but I don't know off-hand how feasible that is.

dsmilkov:
Is your data in HF datasets format, or just raw parquet files? Also we are in slack room with HuggingFace so I just asked for their advice on how to best load a single shard, and ideally a random sample of many shards. Will follow up shortly.

dsmilkov:
Also I'm assuming the s3 buckets are private? We are using duckdb for really fast parquet reads and will use it for sampling. They support S3 reads directly: https://duckdb.org/docs/guides/import/s3_import.html

You will just need to set some env variables to authorize with S3:
```
S3_REGION=
S3_ENDPOINT=
S3_ACCESS_KEY=
S3_SECRET_KEY=  
```

Let me know if this would work for you

jtalmi:
We are using GCP actually! And we have our own custom datasets in parquet format usually (some .jsonl, .csv, but parquet the main one).

Does duckdb enable smart sampling parquet files within s3/gcs buckets, so that the sample is taken from many shards?

dsmilkov:
I just sent https://github.com/lilacai/lilac/pull/778 for review. In summary, when you have many shards, and you don't want to fully shuffle a dataset, we fetch a few rows from each shard by making a reader each for each shard and flipping a coin to select the next reader.

```
index = random.randint(0, len(readers) - 1)
reader = self._readers[index]
batch = reader.read_next_batch() # Fetch a batch of rows.
```

[...]

PR #773, opened 2023-10-20
Add sampling to our `ParquetSource`

dsmilkov:
- Add sampling for parquet files via duckdb
- Improve memory usage by reading smaller batches from huge parquet files
- Add support for parquet files on S3

This isn't yet addressing the problem of having large number of shards and thus needing to sample from random shards, but that's a follow up.

Towards https://github.com/lilacai/lilac/issues/770


lilac/sources/parquet_source.py
```diff
@@ -22,17 +24,49 @@ class ParquetSource(Source):
...
+    s3_filepaths = [path.replace('gs://', 's3://') for path in filepaths]
+
+    # NOTE: We use duckdb here to increase parallelism for multiple files.
+    sample_suffix = f'USING SAMPLE {self.sample_size}' if self.sample_size else ''
+    self._con.execute(f"""
+      CREATE VIEW t as (SELECT * FROM read_parquet({s3_filepaths}) {sample_suffix});
+    """)
+    res = self._con.execute('SELECT COUNT(*) FROM t').fetchone()
+    num_items = cast(tuple[int], res)[0]
+    self._reader = self._con.execute('SELECT * from t').fetch_record_batch(rows_per_batch=10_000)
```
brilee, line 66: these computations seem to belong in process(), not in setup()

brilee (approved):

lilac/sources/parquet_source.py
dsmilkov: great q.

```
self._reader = self._con.execute('SELECT * from t').fetch_record_batch(rows_per_batch=10_000)
```
returns a lazy iterator, so no data is being read yet, but we found that executing this in setup catches a lot of "setup" bugs like file not found, unrecognized parquet format (broken head), unauthorized S3/GCS bucket read etc.

In addition to this, once you have a reader , you can read the inferred schema before reading the data, and our sources need the schema before process() so they can setup buffer parquet writes ahead of time.

.env
```diff
@@ -19,6 +19,10 @@ DUCKDB_USE_VIEWS=0
 # GCS_REGION=
 # GCS_ACCESS_KEY=
 # GCS_SECRET_KEY=
+# S3_REGION=
```
nsthorat, line 22: add these to env.py instead, which will get documented in https://lilacml.com/api_reference/index.html#lilac.LilacEnvironment

lilac/sources/json_source.py
```diff
@@ -62,7 +61,7 @@ def setup(self) -> None:
   @override
   def source_schema(self) -> SourceSchema:
```
nsthorat, line 62: not for now but maybe we should make schemas optional and let duckdb infer types to reduce cognitive overhead of both sources and signals

then signals are very close to a map

lilac/sources/parquet_source.py
nsthorat, line 66: 10k to a constant?

nsthorat (approved):

.env
dsmilkov: done

lilac/sources/parquet_source.py
dsmilkov: done

lilac/sources/json_source.py
dsmilkov: it's our pq.ParquetWriter that needs a schema ahead of time to setup a writer, before writing a single row to disk. And that schema needs to be consistent with 100% of the rows that are going in that writer to avoid write error. That means we need to see the entire data in order to correctly infer the schema, if not provided by the user. Or we circumvent our writer and get duckdb to read the format and dump to paquet directly.

dsmilkov: note that pq.ParquetWriter also doesn't store everythign in disk, it dumbs to parquet every 128MB row_group_buffer_size with 10k items per rowgroup.

PR #775, opened 2023-10-20
Improve the Export Dataset modal dialog

dsmilkov:
Add support for label filtering in the Download modal UI

DEMO: https://huggingface.co/spaces/lilacai/daniel_staging

https://github.com/lilacai/lilac/assets/2294279/1bf67395-c75f-43f9-80dd-eab23f51d006

Towards https://github.com/lilacai/lilac/issues/761


lilac/router_dataset.py
```diff
@@ -309,6 +308,41 @@ def get_media(namespace: str, dataset_name: str, item_id: str, leaf_path: str) -
...
+
+@router.get('/serve_dataset')
+def serve_dataset_file(filepath: str) -> FileResponse:
+  """Serve the exported dataset file."""
+  filepath = os.path.expanduser(filepath)
+  return FileResponse(filepath)
+
+
+@router.post('/{namespace}/{dataset_name}/download')
+def download_dataset(namespace: str, dataset_name: str, options: DownloadOptions) -> str:
```
nsthorat, line 329: should this be called export_dataset since this is returning a filepath?

web/blueprint/src/lib/components/datasetView/DownloadModal.svelte
```diff
@@ -59,81 +78,183 @@
...
+        <div>
+          <RadioButtonGroup bind:selected={selectedFormat}>
+            {#each formats as format}
+              <RadioButton labelText={format} value={format} />
+            {/each}
+          </RadioButtonGroup>
+        </div>
+        <div class="mt-4 max-w-lg pt-2">
+          <TextInput
+            labelText="Output path"
```
nsthorat, line 181: Maybe output filepath? since we use "path" in other places

also might be good to clarify that you need to specify the filename or something -- at first glimpse im not sure if i should put a filename in it or just the dir name

web/blueprint/src/lib/components/datasetView/DownloadModal.svelte
```diff
@@ -59,81 +78,183 @@
...
+            <svelte:fragment slot="actions">
+              <NotificationActionButton on:click={() => window.open(downloadUrl(), '_blank')}>
+                Download
+              </NotificationActionButton>
+            </svelte:fragment>
+          </InlineNotification>
+        {/if}
+      </section>
+      <section>
+        <h2>Step {labelFields.length > 0 ? 4 : 3}: JSON preview</h2>
```
nsthorat, line 214: can we put a little description here saying that this preview is not actually what the file looks like? from the UI I click CSV and it still has json output, slightly non obvious

nsthorat (approved):
wooo nice work

lilac/router_dataset.py
dsmilkov: done

web/blueprint/src/lib/components/datasetView/DownloadModal.svelte
dsmilkov: done

web/blueprint/src/lib/components/datasetView/DownloadModal.svelte
dsmilkov: done

brilee:
To confirm the behavior of the filter - I imagine the way it should work is:
- if any include is specified, then start from the null set and include all things where the include column = True, in a UNION fashion
- otherwise, the include set is the entire dataset by default
- then, apply any exclusion filters, in a UNION fashion (a datapoint is excluded if any of the exclude criteria are met)

PR #777, opened 2023-10-20
Upgrade ruff version and fix associated new linter errors

brilee:


pyproject.toml
```diff
@@ -80,6 +80,7 @@ llama-index = { version = "^0.8.38", optional = true }
 
 # For HDBScan to reduce dimensionality before running clustering.
 umap-learn = { version = "^0.5.4" }
+ruff = "^0.1.1"
```
dsmilkov, line 83: the ruff below in the file is good since it's in the dev dependency section, but let's remove this one since it's not a regular (runtime) dep of lilac

dsmilkov (approved):
lgtm, 1 small comment

nsthorat (approved):
All of the deps got updated so I think it's worthwhile to stand up a demo to make sure nothing broke in the process (since you've updated things like http libraries)

Could be a good time for you to setup huggingface for demos (e.g. our borg).

Details on how to set up the repo here:
https://github.com/lilacai/lilac/blob/main/development.md#demos-for-prs

You'll need a huggingface account if you don't already have one, we can add you to lilacai for billing etc

brilee:
reverted most of the poetry lock changes; diff only includes the ruff upgrade now!

nsthorat:
Nice!! Shipit

PR #791, opened 2023-10-24
Swap in new embedding-chunked splitter algorithm

brilee:
The embedding calculation turned out to be pretty slow - it runs on every sentence, which overall took ~7x longer than calculating embeddings on the final chunks. So the total runtime for computing embeddings got 8x slower.

Instead of using the general purpose embedders, I'm repalcing it with a super cheap and simple bag of words embedding, computed over openai's tokenizer tokens. The new embedder runs ~30x faster than even GTE-tiny, so the total runtime for computing embeddings is only ~5-10% slower than before. 

The BoW vector is folded into a 128-vector, which is plain and simple for now. I haven't gotten the UI working yet so I don't know yet how much degradation we're getting in the chunking quality. The vector dimension is an obvious place to fine-tune the approach later on if we need to. The BoW embedding should at least pick up on things like code vs. text vs. instructions.


dsmilkov (approved):
pretty cool -- very curious to play with the results

brilee:
to minimize the accruing merge conflict debt, I'm going to merge the parts of this PR that don't actually switch the active chunker, and separately figure out why the swap-in is causing the mysterious bug.

PR #792, opened 2023-10-25
Fix a flaky test

dsmilkov:
- Fix a flaky test with multi shard parquet files (it proves that duckdb is parallelizing shard reads)
- Also add `pytest-repeat` plugin as a dev dep so you can run `pytest --count=100`
- Add a couple of code snippets as negative examples to the profanity concept


nsthorat (approved):
Code is profane! Kidding..

brilee (approved):

PR #794, opened 2023-10-25
Add pdb-compatible script and documentation

brilee:


run_server_pdb.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+export LILAC_PROJECT_DIR='./data'
+poetry run python -c "import lilac.server; import uvicorn; uvicorn.run(lilac.server.app, host='0.0.0.0', port=5173)"
```
nsthorat, line 4: s/5173/5432

5173 is actually the vite server, and forwards to the python server on 5432

nsthorat (approved):

run_server_pdb.sh
brilee: intent is to allow a CURL command to work out of the box, so this server should listen to 5173

PR #795, opened 2023-10-26
Improve ergonomics of map().

nsthorat:
- Add the ability to remove the output column so you can just print the output of the map
- Add `overwrite` which overwrites an existing column.
- Ensure that the return value of the function is an iterator of the generator, assuring that the generator is called internally so errors are thrown and map() is called aggressively before returning.


lilac/data/dataset.py
```diff
@@ -68,6 +68,9 @@ def __init__(self, df: pd.DataFrame, total_num_rows: int) -> None:
   def __iter__(self) -> Iterator:
     return (row.to_dict() for _, row in self._df.iterrows())
 
+  def __next__(self) -> Item:
+    return self.__iter__().__next__()
```
dsmilkov, line 72: this looks very strange because calling next 3 times will always return the very first row. It feel like we should either only implement __iter__ (which means SelectRowsResults is iterable), or only implement __next__ (which means SelectRowsResults is an iterator that can be exhausted), but not both.

lilac/data/dataset_duckdb.py
```diff
@@ -1972,34 +1996,44 @@ def _rows() -> Iterable[Item]:
...
+
+    def _generate_output() -> Generator[Item, None, None]:
+      nonlocal reader
+      for batch in reader:
+        yield from (row[output_column] for row in batch.to_pylist())
+      reader.close()
+
+    # Return an iterator from the generator so that the outputs are written before returning a new
+    # iterable.
+    return iter(_generate_output())
```
dsmilkov, line 2036: the pydoc contract in dataset.py says that maps returns an iterable (which I really like), however the impl is returning an iterator, which is less nice to work with since you are only iterator over the data once.

Can you change the result to an iterator instead and add coverage where we iterate over the entire output of dataset.map twice ?

dsmilkov (approved):
great start!

notebooks/DatasetMap.ipynb
```diff
@@ -0,0 +1,159 @@
+{
```
dsmilkov, line 1: <p>Remove the "Query took Xs " output from the cells here and below?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lilacai/lilac/pull/795/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

dsmilkov, line 1: <p>how about <code>itertools.islice(my_list, 3)</code> # grab the first 3 elements</p><p> <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lilacai/lilac/pull/795/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

dsmilkov, line 1: <pre class="ql-syntax" spellcheck="false">Line #10.    print(next(rows))
</pre><p>calling next twice will output the first row twice (see previous comment on next vs iter)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lilacai/lilac/pull/795/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

lilac/data/dataset.py
nsthorat: Done,

lilac/data/dataset_duckdb.py
nsthorat: Done. Huge caveats here -- the iter should open a connection, and close it. Right now its assumed to be opened above, and then it closes after the first iter is exhausted. I have to do this because we store the jsonl in memory, and so the iterator doesnt own it.

In the next PR, when I write to JSONL on disk, I'll clean this all up.

notebooks/DatasetMap.ipynb
nsthorat: Done

nsthorat: I didnt wanna introduce a complicated itertoos, I just iterate now with a for loop

PR #807, opened 2023-10-30
Fix topk on an indexed repeated field + metadata filter

dsmilkov:
The bug shows up when we do a topk search on an indexed repeated field along with metadata filters.


lilac/embeddings/vector_store_hnsw.py
```diff
@@ -120,9 +120,14 @@ def filter_func(label: int) -> bool:
...
     except RuntimeError:
-      # If K is too large compared to M and construction-time ef, HNSW will throw an error.
-      # In this case we return no results, which is ok for the caller of this method (VectorIndex).
-      return []
+      # If K is too large compared to M and construction-time ef, HNSW will throw an error. This
+      # also happens if we are filtering by a tiny subset of keys, where the vector store index
+      # is not able to find enough neighbors. In this case we return dummy results, which is ok
+      # for the caller of this method (VectorIndex).
+      if keys:
+        return [(key, 0.5) for key in keys]
```
nsthorat, line 128: isn't this technically incorrect? Imagine if k=5, and one result shows up in top-k

Why not just remove the filter_func here, and look for keys

nsthorat (approved):

lilac/embeddings/vector_store_hnsw.py
dsmilkov: talked offline. reverted this change. (for posterity, if k=5, there should always be 5 results in the topk since we by sort by continuous float number.

PR #808, opened 2023-10-31
Add JSONL caching for dataset.map().

nsthorat:
This means if your map() throws half way through, it will continue the next time, unless overwrite=True.


lilac/data/dataset_duckdb.py
```diff
@@ -1934,10 +1942,33 @@ def map(
...
     def _rows() -> Iterable[Item]:
-      nonlocal select_queries
+      nonlocal select_queries, con
+
+      # Anti-join removes rows that are already in the cache.
+      anti_join = ''
+      if use_jsonl_cache:
+        anti_join = f"""
+          ANTI JOIN (
+            SELECT * FROM read_json_auto('{jsonl_cache_filepath}', IGNORE_ERRORS=true)) as cache
```
dsmilkov, line 1966: add format='newline_delimited' after ignore_errors so duckdb doesn't have to do guessing on jsonl vs json

lilac/data/dataset_duckdb.py
```diff
@@ -23,6 +23,7 @@
 import pandas as pd
 import pyarrow as pa
 import yaml
+from datasets import fingerprint
```
dsmilkov, line 26: is this used?

lilac/data/dataset_duckdb.py
```diff
@@ -1993,7 +2042,7 @@ def _rows() -> Iterable[Item]:
     tmp_con.execute(f"""
```
dsmilkov, line 2042: do we need to call file.flush() here after we dumps json to disk, in case there is a buffer

dsmilkov (approved):

lilac/data/dataset_duckdb.py
nsthorat: Nope, autoformatter doesnt remove them anymore?

lilac/data/dataset_duckdb.py
nsthorat: Done

lilac/data/dataset_duckdb.py
nsthorat: Shouldn't need to since it's a context manager

PR #826, opened 2023-11-06
Add backend support for sampling jsonl files

brilee:


lilac/sources/json_source.py
```diff
@@ -51,8 +58,11 @@ def setup(self) -> None:
     """
     )
 
-    res = self._con.execute('SELECT COUNT(*) FROM t').fetchone()
-    num_items = cast(tuple[int], res)[0]
+    if self.sample_size:
+      num_items = self.sample_size
+    else:
+      res = self._con.execute('SELECT COUNT(*) FROM t').fetchone()
+      num_items = cast(tuple[int], res)[0]
```
dsmilkov, line 65: do `num_items = min(num_items, sample_size)` in case the sample size is larger than the actual dataset? I guess we should always check the dataset size

lilac/sources/json_source.py
```diff
@@ -51,8 +58,11 @@ def setup(self) -> None:
     """
     )
 
-    res = self._con.execute('SELECT COUNT(*) FROM t').fetchone()
-    num_items = cast(tuple[int], res)[0]
+    if self.sample_size:
+      num_items = self.sample_size
```
nsthorat, line 62: what happens if the sample_size happens to be larger than the dataset?

maybe we should always count, and then max()

lilac/sources/json_source.py
```diff
@@ -71,8 +81,11 @@ def process(self) -> Iterable[Item]:
     if not self._reader or not self._con:
       raise RuntimeError('JSON source is not initialized.')
 
-    for batch in self._reader:
-      yield from batch.to_pylist()
+    items: Iterator[Any] = itertools.chain.from_iterable(
```
dsmilkov, line 84: i think we can get duckdb to do this work for us using provide a SAMPLE clause in the SQL: https://duckdb.org/docs/sql/samples.html

dsmilkov (approved):

lilac/sources/json_source.py
```diff
@@ -51,8 +58,11 @@ def setup(self) -> None:
...
 
-    res = self._con.execute('SELECT COUNT(*) FROM t').fetchone()
-    num_items = cast(tuple[int], res)[0]
+    if self.sample_size:
+      num_items = self.sample_size
+    else:
+      res = self._con.execute('SELECT COUNT(*) FROM t').fetchone()
+      num_items = cast(tuple[int], res)[0]
 
     self._reader = self._con.execute('SELECT * from t').fetch_record_batch(rows_per_batch=10_000)
```
nsthorat, line 67: maybe do a limit in SQL in duckdb instead to short circuit earlier and keep RAM lower

nsthorat (approved):
nice, knocking them off!

brilee:
Confirmed that the option is exposed in UI by default, no special effort needed!

Issue #707, opened 2023-09-22
Changing preferred settings breaks the search panel when there's an embedding computed.


nsthorat:
When gte-small is my default, everything looks good:

<img width="1040" alt="image" src="https://github.com/lilacai/lilac/assets/1100749/9131568c-e171-4d74-b9c0-459a7e0a9b50">

Steps to repro:

Go to settings, change default embedding:
<img width="748" alt="image" src="https://github.com/lilacai/lilac/assets/1100749/eee7f29d-6465-425e-aab2-37dec2dc42c0">

Go back to openorca, dropdown doesnt work:

<img width="1069" alt="image" src="https://github.com/lilacai/lilac/assets/1100749/4c5e8b9d-8259-4ab3-a3e0-35815016832c">

brilee:
the underlying problem is that we don't use existing embeddings if they don't match the settings; we insist on using the configured embedding.

PR #833, opened 2023-11-07
Upgrade Cohere embeddings to v3-light

brilee:


brilee:
https://txt.cohere.com/introducing-embed-v3/ some subtlety around "input_type" - I chose the one they said was their general purpose one. For RAG-style interactive "ask a question and find relevant answers from the dataset" use cases, a different input_type is recommended. Perhaps we should register two different embedders, one general purpose and one for RAG?

brilee:
tried digging into cohere embeddings search (using the query embedder for search queries only) but ran into https://github.com/lilacai/lilac/issues/707

nsthorat (approved):

dsmilkov (approved):

PR #839, opened 2023-11-08
Integrate Presidio into PII detection

### LINKED ISSUE #786: Improve our email regex to avoid emails starting with backtick
We should avoid things like \`@app.route and \`@pytest.fixture as emails.

CleanShot 2023-10-24 at 08 44 44@2x

### LINKED ISSUE #825: Use presidio for PII detection.
https://github.com/microsoft/presidio/

This will greatly improve the PII signal, with first-class support for anonymization and customization.

brilee:
Fixes #825. Fixes #786.


nsthorat (approved):
Amazing work!

For these, can you spin up a demo in the future so we can feel them out (should be instructions in development.md)? I checked out your branch, and ran it, it looks great

Can we convert these upper cases to lower case for consistency with other signals (an explicit dictionary)?

image

brilee:
Space successfully uploaded at https://huggingface.co/spaces/lilacai/brian_staging

The phone number had a high number of false positives; after setting a threshold at 0.5, the FPs are reduced. (if I had to eyeball numbers, recall=1, precision=0.05 before the threshold; recall=0.5, precision=0.5 after the thresholding.) Either the spacy model doesn't support good phone number accuracy, or Presidio's phoen number detector is meh.

Issue #821, opened 2023-11-04
Add column-level diffing.


nsthorat:
For now, we could let the user choose this from the UI.

PR #845, opened 2023-11-11
Add monaco and enable column-level diffing.

nsthorat:
- Add monaco for viewing data like code, and diffing.
- Add a <ButtonDropdown> component for a button that opens a ComboBox since we now use it in 3 places.
- Add state to the URL

Fixes https://github.com/lilacai/lilac/issues/821

https://lilacai-nikhil-staging.hf.space/datasets#lilac/OpenOrca-100k&compareColumns=%5B%7B%22column%22%3A%5B%22response%22%5D%2C%22compareToColumn%22%3A%5B%22question%22%5D%2C%22swapDirection%22%3Afalse%7D%5D

https://github.com/lilacai/lilac/assets/1100749/bb810414-6e6c-4f12-84d3-189359e5fc18

### DIFF
--- a/web/blueprint/src/lib/view_utils.ts
+++ b/web/blueprint/src/lib/view_utils.ts
@@ -69,7 +69,7 @@ export const DTYPE_TO_ICON: Record<DataType, typeof CarbonIcon> = {
   null: NotAvailable
 };
 
-export const SIDEBAR_TRANSITION_TIME_MS = 250;
+export const SIDEBAR_TRANSITION_TIME_MS = 150;
 
 export function getHighlightedFields(
   selectRowsOptions: SelectRowsOptions,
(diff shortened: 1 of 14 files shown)


web/blueprint/src/lib/view_utils.ts
```diff
@@ -69,6 +69,8 @@ export const DTYPE_TO_ICON: Record<DataType, typeof CarbonIcon> = {
   null: NotAvailable
 };
 
+export const SIDEBAR_TRANSITION_TIME_MS = 250;
```
dsmilkov, line 72: make this is a little faster to avoid a few text re-layouts during transition. maybe 150

dsmilkov (approved):
awesome work! how big is the dep?

web/blueprint/src/lib/view_utils.ts
nsthorat: Done!

nsthorat:
~17K over the wire for the editor, but definitely worth it since I think we should switch everything over.

nsthorat:
And also, it's fetched separately, so will get heavily cached.

PR #846, opened 2023-11-12
Add Single Item as a view type with pagination

dsmilkov:
Work mostly done on the plane ride on Friday: https://huggingface.co/spaces/lilacai/daniel_staging

https://github.com/lilacai/lilac/assets/2294279/df32a111-f1ee-4816-8b02-54bf9bc8c8c2


web/blueprint/src/lib/components/datasetView/Dataset.svelte
```diff
@@ -148,7 +152,13 @@
     >
       <SchemaView />
     </div>
-    <div class="h-full w-2/3 flex-grow"><RowView /></div>
+    <div class="h-full w-2/3 flex-grow">
+      {#if viewType == 'scroll'}
+        <RowView />
```
nsthorat, line 157: worth renaming this to ScrollView?

web/blueprint/src/lib/components/datasetView/SingleItemView.svelte
```diff
@@ -0,0 +1,137 @@
...
+  $: rowsQuery = querySelectRows(
+    $store.namespace,
+    $store.datasetName,
+    {...selectOptions, columns: [ROWID], limit},
+    $selectRowsSchema.data?.schema
+  );
+
+  $: manifest = queryDatasetManifest($store.namespace, $store.datasetName);
+
+  // Reset the index to 0 if the row id is not set.
```
nsthorat, line 41: might be worth a top-level comment explaining what this is all for, e.g. "find the rowid by progressively loading the query"

web/blueprint/src/lib/components/datasetView/SingleItemView.svelte
```diff
@@ -0,0 +1,137 @@
...
+    if (index == null) {
+      return;
+    }
+    // Unset the row id and set the new index.
+    store.setRowId(null);
+    index = next ? index + 1 : Math.max(index - 1, 0);
+  }
+</script>
+
+<FilterPanel totalNumRows={$rowsQuery?.data?.total_num_rows} manifest={$manifest.data} />
```
nsthorat, line 88: maybe this should be separated into 2 components -- one that is unconditionally rendered on-top of <SingleItemView> or <ScrollView>, and another that is for your header stuff for this view

web/blueprint/src/lib/components/datasetView/StringSpanHighlight.svelte
```diff
@@ -52,6 +52,10 @@
 
   // When defined, enables semantic search on spans.
   export let datasetViewStore: DatasetViewStore | undefined = undefined;
+  export let alwaysExpand = false;
+
+  let isExpanded = alwaysExpand;
```
nsthorat, line 57: can you add this logic to ItemMediaDiff where I do Math.min(MAX_MONACO_HEIGHT_PX, ...) and remove this max height when you're in this mode? I do this just for the scroll view, but feels like what you're doing here where we should let it be it's natural height. If that feels weird, maybe just increase the max in this view

web/blueprint/src/lib/stores/datasetViewStore.ts
```diff
@@ -47,6 +47,11 @@ export interface DatasetViewState {
   // View.
   schemaCollapsed: boolean;
   insightsOpen: boolean;
+
+  // Currently selected rowid. null is a temp state of loading the next page of row ids to find
+  // the next row id.
```
nsthorat, line 52: maybe specify what undefined means if they're different

nsthorat (approved):
MASSIVE LGTM. Nice work, this is super slick.

One conceptual thing (this is definitely for a follow up): we can probably pre-load metadata for the chunk of ids together (instead of 1 by 1), relying on batshit to batch them together (so the cache key is still per-id). This solves two issues:
1. When I click next, it's slow to load that single id
2. if I hit next like 20 times in a row, it's *extremely* slow as it loads 1 by 1 (and batshit doesn't kick in here because the time between them is > 10ms)

I think there's just a huge amount of overhead for a single request, so whatever we can do to batch will be a win.

Few UI things:
- Can you make arrow keys work now? It feels very natural
- You probably can remove "Item" here, it's clear:

image

- Can we add some margins around the l/r buttons:

image

brilee:
How do we handle deep linking? Is that a separate followup PR?

IMO we shouldn't push onto browser history - that will quickly exhaust the history stack. But we should provide an easy click-to-copy-deeplink-to-clipboard feature.

web/blueprint/src/lib/components/datasetView/Dataset.svelte
dsmilkov: done

dsmilkov:
Deeplinking is already there (done automatically by our dataset view store) and Nikhil mentioned there is no problem with the history stack, browsers can handle a lot

web/blueprint/src/lib/components/datasetView/SingleItemView.svelte
dsmilkov: will do in a follow up

PR #847, opened 2023-11-13
Add parallelism to dataset.map with dask.

nsthorat:
dataset.map() now takes a num_jobs argument for splitting the work. When passed -1, it will use all the cores on the machine.

- We now automatically create a set of tasks for the user when calling dataset.map(), showing up in the UI automatically. If you ask for N jobs, N tasks are created in the UI. To avoid over-generalization, I decided not to push this into tasks for now.
- When work is split between multiple jobs, each worker gets chunk offsets and requests the data independently, and writes to shards independently. The scheduler is responsible for joining all the jsonl files to parquet at the very end.
- When you call map() with no output, we now write to disk to allow communication between workers and the caller.
- Lots of unit tests to make sure continuation still works.

Other things I had to do:
- Add nest_asyncio, which patches asyncio to have event loops allowed to be created in a nested fasion, which enables all the asyncio stuff from jyupter notebooks.
- Lots of fixes to tasks.py to make it propagate errors.
- Adds a TestDaskLogger for use in unit tests where we want to communicate between workers and the scheduler without dealing with dask api.
- Remove freeze_time from tests that interact with dask. This completely freezes the unit tests.
- PII tests seemed to have been checked in in half-baked state, since we have confidence scores but dont assert them in unit tests.


lilac/tasks.py
```diff
@@ -33,6 +36,16 @@
 from .env import env
 from .utils import log, pretty_timedelta
 
+# nest-asyncio is used to patch asyncio to allow nested event loops. This is required when Lilac is
+# run from a Jyupter notebook.
+# https://stackoverflow.com/questions/46827007/runtimeerror-this-event-loop-is-already-running-in-python
+try:
+  # Check if in an iPython environment, then apply nest_asyncio.
+  eval('__IPYTHON__')
```
brilee, line 44: shouldn't need the eval..
`if '__IPYTHON__' in locals():`

dsmilkov (approved):
pretty cool!

lilac/tasks.py
nsthorat: Done! Nice!

lilac/data/dataset_test_utils.py
```diff
@@ -105,3 +106,25 @@ def make_vector_index(
...
   return vector_index
+
+
+class TestDaskLogger:
+  """A simple utility for unit tests that need logging between dask workers and the scheduler."""
+
+  def __init__(self, tmp_dir: pathlib.Path):
+    self.tmp_dir = tmp_dir
+
+  def log_event(self, value: Any) -> None:
```
brilee, line 117: I don't think the "open/close each file" works to prevent race conditions. You'll need a threading.Lock() owned by the TestDaskLogger class, and use `with lock:` block to protect log_event, clear_logs, and get_logs

lilac/tasks.py
```diff
@@ -244,10 +292,16 @@ def get_is_dask_worker() -> bool:
     return False
 
 
-@functools.cache
```
brilee, line 247: curious - is there a reason this didn't work?

lilac/data/dataset_duckdb.py
```diff
@@ -2049,18 +2068,146 @@ def map(
...
         else:
           raise ValueError(
             f'Cannot map to path "{output_path}" which already exists in the dataset.'
           )
     else:
       output_column = map_fn.__name__
 
+    num_items = self.manifest().num_items + 1
+    num_workers = get_task_manager().get_num_workers() if num_jobs == -1 else num_jobs
+    worker_chunk_size = math.ceil(num_items / num_workers)
```
brilee, line 2096: https://numpy.org/doc/stable/reference/generated/numpy.array_split.html#numpy.array_split

nsthorat: Ahhhh cant use this, arrays!

PR #848, opened 2023-11-13
Pass the job_id to the dataset.map map_fn.

nsthorat:


lilac/schema.py
```diff
@@ -132,6 +132,23 @@ def signal_type_supports_dtype(input_type: SignalInputType, dtype: DataType) ->
...
+
+  __name__: str
+
+  def __call__(self, row: Item, job_id: int) -> Item:
+    """Calls the map function, mapping an item to an item.
+
+    Argumnets:
+      row: An Item, a dictionary of an entire row.
+      job_id: The job id.
+      *args: Additional arguments.
```
brilee, line 146: the args and kwargs aren't in the signature of __call__? Not sure if that matters for Protocol.

lilac/schema.py
```diff
@@ -132,6 +132,23 @@ def signal_type_supports_dtype(input_type: SignalInputType, dtype: DataType) ->
 Bin = tuple[str, Optional[Union[float, int]], Optional[Union[float, int]]]
 
 
+class MapFn(Protocol):
+  """Interface for a map function."""
+
+  __name__: str
+
+  def __call__(self, row: Item, job_id: int) -> Item:
```
brilee, line 140: Wondering about the map_fn knowing its job_id - I guess the expectation is that a shard can vary their behavior based on which job_id they are? It breaks the mapping abstraction a bit and I don't see how it might be used.

brilee (approved):

lilac/schema.py
nsthorat: Correct -- this is for a customer that wants to set num_jobs=num_proc, and then assign a GPU to each process

PR #858, opened 2023-11-16
Add jsonl intermediate caching to signals. Introduce a central spot for this cache abstraction.

nsthorat:
### `_compute_disk_cached`
Computes a signal, or map function, over a selected input. Every output is saved to disk via jsonl, and can be sharded amongst multiple processes.

### `_reshard_cache`
Re-shards the jsonl files into a parquet file with duckdb.

These two abstractions:
- Allow restarting based on the cache-key of the output path by anti-joining to the jsonl files. We assume here that shards are consistent (e.g. the select sorts by ROWID)
- Compute progress via the UI.
- Allow explicitly overwriting (will throw if it exists)
- Embedding computation does *not* get cached via jsonl. It get cached as it previously did. This means if you run an embedding again, there are no restarts. We can fix this later.

These two are now used for Dataset.map() and Dataset.compute_signal().

Other changes:
- Instead of ranges, shard indices and counts are sent to workers of map
- I removed the column selection for dataset.map(), it always selects everything. This is kind of confusing, and and over-generalization.

Side-effects:
- There's a funky side-effect where signals now can emit float32, but on the other side of select_rows it will be float64 because of going through jsonl. This can be solved by explicitly passing a schema to the COPY {json} TO PARQUET but is non-trivial because of nested structs.

As a side-effect, signals also can now have optional schemas. I didn't do this yet, but I'll do it in a follow up to keep this smaller.

https://github.com/lilacai/lilac/assets/1100749/00e43943-8dbe-4d5f-ad2b-1e8c7150ecbd


lilac/data/dataset_compute_signal_test.py
```diff
@@ -196,17 +197,6 @@ def setup_teardown() -> Iterable[None]:
   clear_signal_registry()
 
 
-def test_signal_output_validation(make_test_data: TestDataMaker) -> None:
```
dsmilkov, line 199: can we keep this test, in the case where no errors were thrown, so we can protect the signal authors from a fan-out?

lilac/schema.py
```diff
@@ -177,8 +177,6 @@ def validate_fields(cls, fields: Optional[dict[str, 'Field']]) -> Optional[dict[
     """Validate the fields."""
     if not fields:
       return fields
-    if SPAN_KEY in fields:
```
dsmilkov, line 180: curious what changed so this needs to go away?

lilac/router_dataset.py
```diff
@@ -113,7 +113,13 @@ def _task_compute_signal(
     # underlying pydantic serializer.
     options = ComputeSignalOptions(**options_dict)
     dataset = get_dataset(namespace, dataset_name)
-    dataset.compute_signal(options.signal, options.leaf_path, task_step_id=(task_id, 0))
+    dataset.compute_signal(
```
dsmilkov, line 116: for simpler Rest API, should we add /compute_embedding endpoint, analogous to dataset.compute_embedding?

lilac/load.py
```diff
@@ -263,7 +263,9 @@ def _compute_signal(
     del os.environ['DEBUG']
 
   dataset = get_dataset(namespace, name, project_dir)
-  dataset.compute_signal(signal_config.signal, signal_config.path, task_step_id)
+  dataset.compute_signal(
+    signal=signal_config.signal, path=signal_config.path, overwrite=True, task_step_id=task_step_id
```
dsmilkov, line 267: should we allow restarts in load? that seems useful. In that case, we can propagate the `overwrite` argument of `load()` above here: `overwrite=overwrite`

lilac/data/dataset.py
```diff
@@ -392,36 +392,46 @@ def update_settings(self, settings: DatasetSettings) -> None:
...
   def compute_embedding(
-    self, embedding: str, path: Path, task_step_id: Optional[TaskStepId] = None
+    self,
+    embedding: str,
+    path: Path,
+    overwrite: Optional[bool] = False,
+    task_step_id: Optional[TaskStepId] = None,
   ) -> None:
     """Compute an embedding for a given field path."""
     signal = get_signal_by_type(embedding, TextEmbeddingSignal)()
```
dsmilkov, line 420: until we make embeddings restartable, might catch intermediate bugs by checking `assert overwrite is False, 'overwrite for embeddings is not yet supported'`

lilac/data/dataset_duckdb.py
```diff
@@ -521,80 +519,321 @@ def _get_vector_db_index(self, embedding: str, path: PathTuple) -> VectorDBIndex
...
+        SELECT * FROM t
+        ORDER BY {ROWID}
+        LIMIT {shard_end_idx - shard_start_idx}
+        OFFSET {shard_start_idx}
+      );
+    """
+    )
+    select_sql = ', '.join(select_queries)
+
+    start_idx = 0
```
dsmilkov, line 629: should we invalidate the cache if for whatever reason dataset.size has changed? to avoid offset errors in shard workers

lilac/data/dataset.py
```diff
@@ -392,36 +392,46 @@ def update_settings(self, settings: DatasetSettings) -> None:
 
   @abc.abstractmethod
   def compute_signal(
-    self, signal: Signal, path: Path, task_step_id: Optional[TaskStepId] = None
+    self,
+    signal: Signal,
+    path: Path,
+    overwrite: Optional[bool] = False,
```
brilee, line 398: the ternary boolean strikes! I think this should just be `bool` :D

lilac/data/dataset.py
nsthorat: The restartability is actually orthogonal, now we just error if it's there so you dont accidentally overwwrite.

lilac/data/dataset.py
```diff
@@ -392,36 +392,46 @@ def update_settings(self, settings: DatasetSettings) -> None:
...
+    task_step_id: Optional[TaskStepId] = None,
   ) -> None:
     """Compute a signal for a column.
 
     Args:
       signal: The signal to compute over the given columns.
       path: The leaf path to compute the signal on.
+      overwrite: Whether to overwrite an existing signal computed at this path.
```
brilee, line 406: In what scenarios would you expect a signal to output a different value the second time? I guess this could happen if the implementation is changed. Or is there something else where an interrupted signal is corrupted somehow?

lilac/data/dataset_utils.py
```diff
@@ -243,6 +244,13 @@ def write_items_to_parquet(
   return out_filename
 
 
+class WriteItemsResponse(BaseModel):
```
brilee, line 247: unused?

lilac/schema.py
nsthorat: Signals can generate span keys, and then we infer the schema from it. Then it yells. This actually isn't a reserved field, only __VALUE__ is, because it's a runtime construct (joining two things). SPAN_KEY is actually written to disk.

lilac/router_dataset.py
nsthorat: Yes, good idea, though will do this later because it's non-trivial UI stuff

lilac/data/dataset_duckdb.py
```diff
@@ -521,80 +519,321 @@ def _get_vector_db_index(self, embedding: str, path: PathTuple) -> VectorDBIndex
...
+    manifest = self.manifest()
+
+    num_items = self.manifest().num_items
+    shard_size = math.ceil(num_items / shard_count) if shard_count else num_items
+    shard_start_idx = shard_id * shard_size if shard_id is not None else 0
+    shard_end_idx = (
+      min((shard_id + 1) * shard_size, num_items) if shard_id is not None else num_items
+    )
+
+    output_path = (
```
brilee, line 544: `output_path = output_path or getattr(transform_fn, 'name') or transform_fn.__name__`

lilac/data/dataset_duckdb.py
nsthorat: Thinking about this more -- no we shouldn't. We should actually join with rowid when we actually can change num_items, to ensure that the rowid actually exists, before writing it to parquet.  This means shards can still be caches (if you add a row, we just compute the rest, if you remove a row, we join to the original table out of jsonl before writing parquet).

This is actually going to be a complicated thing to think through what happens if you change num_items and what we do about signals, so let's solve it them (instead of just slapping num_items on jsonl).

brilee (commented):
This seems unnecessarily complicated / reimplementing task management.

What about a single .sqlite file with a table having two columns: row_id, and output_json?

Then all the cache workers would concurrently update the rows of this sqlite file as they complete it, and the single sqlite file represents the granular completion state of the job.

I think this solution would likely cut down tremendously on the amount of custom duckdb code we have going here...

lilac/data/dataset_duckdb.py
```diff
@@ -521,80 +519,322 @@ def _get_vector_db_index(self, embedding: str, path: PathTuple) -> VectorDBIndex
...
-    spec = _split_path_into_subpaths_of_lists(enriched_path)
+    return jsonl_cache_items
 
-    enriched_signal_items = cast(Iterable[Item], wrap_in_dicts(nested_out, spec))
+  def _reshard_cache(
+    self,
+    output_path: PathTuple,
+    parquet_filename_prefix: Optional[str] = None,
+    shard_count: Optional[int] = None,
+    disable_write_parquet: bool = False,
```
brilee, line 766: this kwarg seems redundant with just testing if output_path is None

lilac/data/dataset_duckdb.py
```diff
@@ -521,80 +519,322 @@ def _get_vector_db_index(self, embedding: str, path: PathTuple) -> VectorDBIndex
...
+      os.makedirs(os.path.dirname(parquet_filepath), exist_ok=True)
+
+      con.execute(
+        f"""COPY (SELECT * FROM "{jsonl_view_name}") TO '{parquet_filepath}' (FORMAT PARQUET);"""
+      )
+
+      con.close()
+
+    output_schema = arrow_schema_to_schema(reader.schema)
+    if ROWID in output_schema.fields:
```
brilee, line 821: is it a problem if there is a ROWID in the reader and/or written parquet, but the schema is missing the rowid?

lilac/data/dataset_duckdb.py
```diff
@@ -2540,21 +2619,63 @@ def _signal_dir(enriched_path: PathTuple) -> str:
...
+) -> str:
+  # The final parquet file is not sharded.
+  parquet_filename = get_parquet_filename(
+    parquet_filename_prefix or output_path[-1], shard_index=0, num_shards=1
+  )
+  path_prefix: Optional[PathTuple]
+  if parquet_filename_prefix:
+    path_prefix = output_path
+  else:
+    path_prefix = output_path[0:-1] if len(output_path) > 1 else None
```
brilee, line 2635: don't need the 0 - could just be `output_path[:-1]`

brilee (approved):

lilac/data/dataset_duckdb.py
nsthorat: this is deleted here because it's going to be merged at the top-level later. this is something ill resolve when i do better auto-inference

nsthorat:
Spoke offline but the complexity isn't actually about just caching it's:
- Merging several parquet files into a single view which pretends that it's all joined in a single table (this is 90% of the complexity)
- Supporting unnested, or dictionary inputs
- Supporting combine_cols, or resolve_spans
- anti-join, avoiding calling map() with cached inputs

The actual cache is like 5ish lines of code, so whether it's sqlite or jsonl won't really change the fact that we have to have to do all this stuff regardless.

PR #866, opened 2023-11-20
Add a nest_under field to dataset.map().

nsthorat:
Changes to `Dataset.map`:
- Adds `nest_under` to allows users to decide to put the output column under another column, allowing annotations like span highlighting from the UI.
- Renames `output_path` to `output_column` as a string to name the containing column for the map.
- Adds automatic inference for {__span__: {start: int, end: int}, 'metadata': ...} so that users dont have to define a schema for map and API calls to ll.lilac_span() will automatically create a span schema for highlighting.

Other *important* changes:
- When calling select_rows with combine_cols=False, we now flatten all columns to top-level. This is the same for the call to dataset.map().

[[Demo](https://lilacai-nikhil-staging.hf.space/datasets#local/glue_ax_map&showMetadataPanel=true&expandedStats=%7B%22premise.thes%22%3Atrue%7D&query=%7B%22filters%22%3A%5B%7B%22path%22%3A%5B%22premise%22%2C%22thes%22%5D%2C%22op%22%3A%22equals%22%2C%22value%22%3A%22the%22%7D%5D%7D)] [[Notebook](https://app.reviewnb.com/lilacai/lilac/pull/866/)]

Code that generates the UI below:
```py
import re
# Find instances of 'the'.
MATCH_REGEX = 'the'


def _find_the(row: dict, job_id: int) -> ll.Item:
  return [ll.lilac_span(m.start(), m.end()) for m in re.finditer(MATCH_REGEX, row['premise'])]

glue.map(
  _find_the, output_column='thes', nest_under='premise', overwrite=True, combine_columns=False
)

# Print the first three rows.
rows = glue.select_rows([ll.PATH_WILDCARD], combine_columns=False, limit=3)
for row in rows:
  print(row)
```

image

### DIFF
--- a/lilac/data/dataset.py
+++ b/lilac/data/dataset.py
@@ -589,7 +589,8 @@ class Dataset(abc.ABC):
   def map(
     self,
     map_fn: MapFn,
-    output_path: Optional[Path] = None,
+    output_column: Optional[str] = None,
+    nest_under: Optional[Path] = None,
     overwrite: bool = False,
     combine_columns: bool = False,
     resolve_span: bool = False,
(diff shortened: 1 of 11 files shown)


dsmilkov (approved):
pretty awesome!

PR #870, opened 2023-11-21
Add `map<key, value>` dtype. No support in the UI yet.

dsmilkov:
This adds support for the `map<key, value>` dtype. The refactor enabled us to add future parameterized dtypes, like `decimal128<precision,scale>`.

This PR doesn't yet add capabilities in the UI where we can compute histogram of the map keys, but it allows ingestion of a parquet source with map.

`DataType` changed from an `enum` to a Pydantic model. Kept back-compat by implementing custom deserialization from string (e.g. "string_span") to a pydantic DataType model.


lilac/schema.py
```diff
@@ -58,47 +59,115 @@
...
+FLOAT16 = DataType('float16')
+FLOAT32 = DataType('float32')
+FLOAT64 = DataType('float64')
+
+### Time ###
+# Time of day (no time zone).
+TIME = DataType('time')
+# Calendar date (year, month, day), no time zone.
+DATE = DataType('date')
+# An "Instant" stored as number of microseconds (µs) since 1970-01-01 00:00:00+00 (UTC time zone).
```
nsthorat, line 161: copilot? maybe just remove lol

nsthorat (approved):
nice

lilac/schema.py
dsmilkov: was there at head, so best to keep it

PR #875, opened 2023-11-21
UI fixes for expanding long rows.

nsthorat:
- Adds an expand button when the media text is too long, both for `StringSpanHighlight` and `ItemMediaDiff`.
- Adds a linear-gradient when text is too long for string span highlight.
- Turn line numbers on in monaco diff editor
- Add a hoist action to hoist the button dropdown combobox so it's on top of other components.
- Fix a bug where media field order kept being reset by the settings panel.

Demo: https://lilacai-nikhil-staging.hf.space/datasets#local/OpenOrca-10k&rowId=%22ff6f272a4e034d459dce9c5b7d5c7761%22&compareColumns=%5B%7B%22column%22%3A%5B%22response_map%22%5D%2C%22compareToColumn%22%3A%5B%22response%22%5D%2C%22swapDirection%22%3Afalse%7D%5D

https://github.com/lilacai/lilac/assets/1100749/97264d67-bbd9-41af-8a5d-beeb8512d774


brilee:
I like it!

Two potential improvements: More hints that there's an expand button, and perhaps a persistent setting or option for opening rows. After all, the use case is probably that a user _always_ wants to see the row expanded, or _never_ wants to see it, depending on what they're inspecting.

dsmilkov (approved):

Issue #876, opened 2023-11-21
Compute signal page is broken -- no fields available.


nsthorat:
Click any column, then compute signal.

Probably from the map PR.

<img width="757" alt="image" src="https://github.com/lilacai/lilac/assets/1100749/9fc5de5a-89d5-4806-9639-a5169bacd8f3">

PR #877, opened 2023-11-21
Fix small bug with compute signal / concepts and filtering by valid dtypes.

nsthorat:
Fixes https://github.com/lilacai/lilac/issues/876

This bug was introduced when the dtype string was put under dtype.type.


brilee (approved):

dsmilkov (approved):
nice find!

PR #880, opened 2023-11-22
Allow custom signals to work with dask processes.

nsthorat:
NOTE: This does not fix Jyupter notebooks which cannot pickle classes defined in it.

We do this by resolving the signal in the scheduler process, and allowing pickle to pickle the entire class, instead of trying to look it up in the registry when the dask worker boots up.

- Upgrade pydantic. New versions support pickling.
- Fix the issue where we swallow errors with "signal outputted a different number of elements than input". Now we separate these and throw if there is a real error, printing the full traceback.


lilac/data/dataset_duckdb.py
```diff
@@ -809,14 +809,20 @@ def _compute_disk_cached(
...
-      try:
+    try:
+      if not isinstance(transform_fn, TextEmbeddingSignal):
         with open_file(jsonl_cache_filepath, 'a') as file:
           for item in output_items:
             json.dump(item, file)
             file.write('\n')
-      except RuntimeError as e:
-        raise ValueError('The signal generated a different number of values than was input.') from e
+    except RuntimeError as e:
```
dsmilkov, line 818: just double checking, how does the mismatch in number of input/outputs result in a runtime error above (where we write to jsonl)?

lilac/router_dataset.py
```diff
@@ -106,15 +106,18 @@ def compute_signal(
   if not get_user_access(user).dataset.compute_signals:
     raise HTTPException(401, 'User does not have access to compute signals over this dataset.')
 
-  def _task_compute_signal(
-    namespace: str, dataset_name: str, options_dict: dict, task_id: TaskId
-  ) -> None:
+  # Resolve the signal outside the task so we don't look up the signal in the registry. This gets
+  # implicitly pickled by the dask serializer when _task_compute_signal is pickled.
+  # NOTE: This unfortunately does not work in Jyupter because a module is not picklable. In this
+  # case, we recommend defining and registering the signal outside a Jyupter notebook.
```
dsmilkov, line 112: typo: jyupter -> jupyter

dsmilkov (approved):
crazy that a single line signal = options.signal made this work

lilac/data/dataset_duckdb.py
nsthorat: Added a comment, this is the exception thrown by the generator when the zipped iterables throws a StopIterator exception

lilac/router_dataset.py
nsthorat: done

PR #881, opened 2023-11-23
Fix a few issues with batching, prefetching, and searches.

nsthorat:
- When you searched by keyword, the state was stale. Now it's updated with the right cache keys in batshit.
- FIx a tiny bug where spans have extra spacing.
-

### DIFF
--- a/web/blueprint/src/lib/queries/datasetQueries.ts
+++ b/web/blueprint/src/lib/queries/datasetQueries.ts
@@ -170,7 +170,14 @@ export const queryRowMetadata = (
 ): CreateQueryResult<Awaited<Record<string, any>>, ApiError> =>
   createApiQuery(
     getRowMetadataBatcher(namespace, datasetName, selectRowsOptions).fetch,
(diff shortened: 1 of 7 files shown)


web/blueprint/src/lib/queries/datasetQueries.ts
```diff
@@ -170,13 +174,21 @@ export const queryRowMetadata = (
 ): CreateQueryResult<Awaited<Record<string, any>>, ApiError> =>
   createApiQuery(
```
dsmilkov, line 175: in this case, since getRowMetadataBatcher returns a function with a very opaque signature, i think we should call `createQuery` directly. Going through our `createApiQuery` adds a step of indirection that is very hard to understand/debug, e.g. hard to figure out the final cache key given the return type of the `getRowMetadataBatcher(...).fetch`

nsthorat: Done, though not sure if it's actually better.

PR #882, opened 2023-11-24
Add support for input paths to dataset.map.

nsthorat:
When passing an input path, the value will get flattened and passed one-by-one to dataset.map. The output structure will parallel the repeated structure of the input.

MapInfo now contains the input_path if passed, so the UI can automatically show lineage information.

Demo (the second copy is generated by a map): https://lilacai-nikhil-staging.hf.space/datasets#local/OpenHermes-2.5

### DIFF
--- a/lilac/data/dataset_map_test.py
+++ b/lilac/data/dataset_map_test.py
@@ -291,6 +291,22 @@ def test_map_input_path_nested(
   ]
 
 
+def test_map_input_path_nonleaf_throws(make_test_data: TestDataMaker) -> None:
+  dataset = make_test_data(
+    [
+      {'id': 0, 'text': ['a']},
+      {'id': 1, 'text': ['b']},
+      {'id': 2, 'text': ['c']},
+    ]
+  )
+
+  def _upper(row: Item, job_id: int) -> Item:
+    return str(row).upper()
+
+  with pytest.raises(Exception):
+    dataset.map(_upper, input_path='text', output_column='text_upper')
+
+
 @pytest.mark.parametrize('num_jobs', [1, 2])
 def test_map_continuation(
   num_jobs: Literal[-1, 1, 2], make_test_data: TestDataMaker, test_dask_logger: TestDaskLogger
--- a/lilac/data/dataset_map_test.py
+++ b/lilac/data/dataset_map_test.py
@@ -218,7 +218,10 @@ def test_map_input_path(num_jobs: Literal[-1, 1, 2], make_test_data: TestDataMak
         'text_upper': field(
           dtype='string',
           map=MapInfo(
-            fn_name='_upper', fn_source=inspect.getsource(_upper), date_created=TEST_TIME
+            fn_name='_upper',
+            input_path=('text',),
+            fn_source=inspect.getsource(_upper),
+            date_created=TEST_TIME,
           ),
         ),
       }
@@ -268,7 +271,10 @@ def test_map_input_path_nested(
         'texts_upper': field(
           fields=['string'],
           map=MapInfo(
-            fn_name='_upper', fn_source=inspect.getsource(_upper), date_created=TEST_TIME
+            fn_name='_upper',
+            input_path=('texts', PATH_WILDCARD, 'value'),
+            fn_source=inspect.getsource(_upper),
+            date_created=TEST_TIME,
           ),
         ),
       }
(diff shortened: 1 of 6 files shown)


lilac/data/dataset_map_test.py
```diff
@@ -191,6 +191,106 @@ def _map_fn(row: Item, job_id: int) -> Item:
...
+def test_map_input_path(num_jobs: Literal[-1, 1, 2], make_test_data: TestDataMaker) -> None:
+  dataset = make_test_data(
+    [
+      {'id': 0, 'text': 'a'},
+      {'id': 1, 'text': 'b'},
+      {'id': 2, 'text': 'c'},
+    ]
+  )
+
+  def _upper(row: Item, job_id: int) -> Item:
```
dsmilkov, line 204: style nit: no need to underline when the fn is internal of a method, or simulates user code (like this unit test). generally underline is to signal that a top level fn that spills into the module members should be private API

lilac/data/dataset_map_test.py
```diff
@@ -191,6 +191,106 @@ def _map_fn(row: Item, job_id: int) -> Item:
...
+    return str(row).upper()
+
+  # Write the output to a new column.
+  dataset.map(_upper, input_path='text', output_column='text_upper', num_jobs=num_jobs)
+
+  # The schema should not reflect the output of the map as it didn't finish.
```
dsmilkov, line 210: update comment since the map did finish

lilac/data/dataset_map_test.py
```diff
@@ -191,6 +191,106 @@ def _map_fn(row: Item, job_id: int) -> Item:
...
+            fn_source=inspect.getsource(_upper),
+            date_created=TEST_TIME,
+          ),
+        ),
+      }
+    ),
+    num_items=3,
+    source=TestSource(),
+  )
+  # The rows should not reflect the output of the unfinished map.
```
dsmilkov, line 232: update comment since the map finished

lilac/data/dataset_map_test.py
```diff
@@ -191,6 +191,106 @@ def _map_fn(row: Item, job_id: int) -> Item:
...
+    return str(row).upper()
+
+  dataset.map(
+    _upper,
+    input_path=('texts', PATH_WILDCARD, 'value'),
+    output_column='texts_upper',
+    num_jobs=num_jobs,
+  )
+
+  # The schema should not reflect the output of the map as it didn't finish.
```
dsmilkov, line 263: same here, and below, the map finished

lilac/data/dataset_map_test.py
```diff
@@ -191,6 +191,106 @@ def _map_fn(row: Item, job_id: int) -> Item:
...
+      {'id': 2, 'texts': [{'value': 'd'}, {'value': 'e'}, {'value': 'f'}]},
+    ]
+  )
+
+  def _upper(row: Item, job_id: int) -> Item:
+    return str(row).upper()
+
+  dataset.map(
+    _upper,
+    input_path=('texts', PATH_WILDCARD, 'value'),
```
dsmilkov, line 258: can you add a test where the input_path is `('texts', PATH_WILDCARD)`, and then the map just does len(x) and there shouldn't be any flattenning, since we are pointing at the array, not at a member of a struct inside that array.

dsmilkov (approved):
awesome!

lilac/data/dataset_map_test.py
dsmilkov: Actually, 2 tests needed:
a) `input_path = ('texts', PATH_WILDCARD)` where we do flatten and the map takes a `struct(value: string)`.
b) `input_path = ('texts',)` where we do not flatten and the map takes a `list(struct(value:string))` and does `len(x)`

lilac/data/dataset_map_test.py
nsthorat: I prefer this to make it clear it's defined only for this test

lilac/data/dataset_map_test.py
nsthorat: Done

lilac/data/dataset_map_test.py
nsthorat: Done

lilac/data/dataset_map_test.py
nsthorat: We now throw if it's not a leaf. Will fix this some time in the future, but not a P0 as it's complicated (deep in nesting hell, and lots of other questions come up about what the shape out the output is when you have nest_under).

lilac/data/dataset_map_test.py
nsthorat: Done

PR #883, opened 2023-11-24
Improve ergonomics of map, relaxing the exact requirement of kwargs={row, job_id}

nsthorat:
The current types & way we call map force users to have exactly "row" and "job_id".

This loosens this requirement with job_id optionally passed only if the function signature contains it.

This allows users to do things like:
```
ds.map(lambda x: x)
```

Allowing them to choose the name of their variable, and not force a job_id.


lilac/data/dataset_duckdb.py
```diff
@@ -2606,9 +2606,34 @@ def _map_worker(
...
+      if has_job_id_kwarg:
+        kwargs['job_id'] = job_id
+      elif has_job_id_arg:
+        args.append(job_id)
+      return args, kwargs
+
     def _map_iterable(items: Iterable[RichData]) -> Iterable[Optional[Item]]:
       for item in items:
-        yield map_fn(item, job_id=job_id)
+        args, kwargs = _get_args_kwargs(item)
```
dsmilkov, line 2635: can you move this outside the for loop to avoid relative slowdown for cheap maps since most of `_get_args_kwargs` doesn't depend on the current item

dsmilkov (approved):
nice, one small comment but important for when map is super cheap (for relative perf)

lilac/data/dataset_map_test.py
```diff
@@ -809,13 +809,112 @@ def _map_fn(row: dict, job_id: int) -> Item:
   ]
 
 
+@pytest.mark.parametrize('num_jobs', [-1, 1, 2])
+def test_map_ergonomics(num_jobs: Literal[-1, 1, 2], make_test_data: TestDataMaker) -> None:
+  dataset = make_test_data([{'text': 'a sentence'}, {'text': 'b sentence'}])
+
+  def _job_fn_kw(item: Item, job_id: int) -> Item:
```
dsmilkov, line 816: i think we should only do positional args (will reduce the overhead at runtime), also align with other libs. If you want to keep the kwargs magic, we should add a test for `map(job_id, item)` where the order is reversed, but i really don't think we need that level of ergo

nsthorat: Done

lilac/data/dataset_duckdb.py
nsthorat: just args now

PR #884, opened 2023-11-25
Upgrade duckdb to 0.9.2, fixing a crash in a dask process with fetch_df_chunk.

nsthorat:


dsmilkov (approved):
awesome, is there a test to repro this? (no worries if repro is complex)

nsthorat:
I couldn't reproduce this outside of a very large dataset unfortunately...

PR #885, opened 2023-11-25
Pass explicit schema during jsonl -> parquet conversion

dsmilkov:
This fixes signal computation where signals like PII.secrets might return empty results. In that case the schema in `signal_manifest.json` didn't match the schema in the `data.parquet` file. 


lilac/schema.py
```diff
@@ -923,3 +923,96 @@ def merge_schemas(schemas: Sequence[Union[Schema, Field]]) -> Schema:
...
+  else:
+    raise ValueError(f'Can not convert dtype "{dtype}" to duckdb type')
+
+
+def _escape_col_name(name: str) -> str:
+  return name.replace('"', '""').replace("'", "''")
+
+
+def _duckdb_field(field: Field) -> str:
+  """Convert a field to a duckdb field."""
```
nsthorat, line 991: maybe _duckdb_struct_type? it's not called a field in duckdb afaict

lilac/schema.py
```diff
@@ -923,3 +923,96 @@ def merge_schemas(schemas: Sequence[Union[Schema, Field]]) -> Schema:
   merged_field = merge_fields([Field(fields=s.fields) for s in schemas])
   assert merged_field.fields, 'Merged field should have fields'
   return Schema(fields=merged_field.fields)
+
+
+def _duckdb_type(dtype: Optional[DataType]) -> str:
+  """Convert the dtype to an duckdb type."""
```
nsthorat, line 929: s/an/a

lilac/schema.py
```diff
@@ -923,3 +923,96 @@ def merge_schemas(schemas: Sequence[Union[Schema, Field]]) -> Schema:
...
+  elif dtype == DATE:
+    return 'DATE'
+  elif dtype == TIMESTAMP:
+    return 'TIMESTAMP'
+  elif dtype == INTERVAL:
+    return 'INTERVAL'
+  elif dtype == EMBEDDING:
+    # We reserve an empty column for embeddings in parquet files so they can be queried.
+    # The values are *not* filled out. If parquet and duckdb support embeddings in the future, we
+    # can set this dtype to the relevan type.
```
nsthorat, line 969: relevant

lilac/schema.py
```diff
@@ -923,3 +923,96 @@ def merge_schemas(schemas: Sequence[Union[Schema, Field]]) -> Schema:
...
+  """Convert a field to a duckdb field."""
+  if field.fields:
+    fields: dict[str, str] = {}
+    for name, subfield in field.fields.items():
+      fields[name] = _duckdb_type(subfield.dtype) if name == ROWID else _duckdb_field(subfield)
+
+    # When nodes have both dtype and children, we add __value__ alongside the fields.
+    if field.dtype:
+      if field.dtype == STRING_SPAN:
+        fields[SPAN_KEY] = f'STRUCT("{TEXT_SPAN_START_FEATURE}" INT, "{TEXT_SPAN_END_FEATURE}" INT)'
```
nsthorat, line 1000: i see this in the method above, should one of them remove it?

lilac/schema.py
```diff
@@ -923,3 +923,96 @@ def merge_schemas(schemas: Sequence[Union[Schema, Field]]) -> Schema:
...
+  elif dtype is None:
+    return 'NULL'
+  elif dtype.type == 'map':
+    map_dtype = cast(MapType, dtype)
+    return f'MAP({_duckdb_type(map_dtype.key_type)}, {_duckdb_type(map_dtype.value_type)})'
+  else:
+    raise ValueError(f'Can not convert dtype "{dtype}" to duckdb type')
+
+
+def _escape_col_name(name: str) -> str:
```
nsthorat, line 986: this method exists in dataset_duckdb too, should we unify?

nsthorat (approved):
wooop nice work, will add this for map too in a follow up

lilac/schema.py
dsmilkov: it's slightly different as in this combines two different escaping methods in one, because of the subtle diff on the signature of the columns arguments in read_json. columns is both a string literal with column literal inside.

nsthorat: gotchya, maybe rename it.

Also, this stuff should actually all move out of schema.py since this file is agnostic of the backend.

lilac/schema.py
dsmilkov: not worth generalizing since there are a bit diferent. this one handles the case when a field is both a pedal (with string_span) and a stuct, so the string_span ends up as a sibling to the other fields. The method above encodes a primitive "string_span", so there is no need to fuse sublings.

lilac/schema.py
dsmilkov: done

lilac/schema.py
dsmilkov: done

lilac/schema.py
dsmilkov: done

lilac/schema.py
dsmilkov: moved out of schema

PR #897, opened 2023-11-29
Add support for `dataset.map()` with entire input as iterable

dsmilkov:


lilac/data/dataset.py
```diff
@@ -593,6 +593,7 @@ def map(
     combine_columns: bool = False,
     resolve_span: bool = False,
     num_jobs: int = 1,
+    entire_input: bool = False,
```
nsthorat, line 596: how about iterable_input?

lilac/data/dataset.py
```diff
@@ -617,6 +618,8 @@ def map(
       num_jobs: The number of jobs to shard the work, defaults to 1. When set to -1, the number of
         jobs will correspond to the number of processors.. If `num_jobs` is greater than the number
         of processors, it split the work into `num_jobs` and distribute amongst processors.
+      entire_input: If true, pass an iterable of the entire input to the map function. When true,
```
nsthorat, line 621: "If true, pass an iterable of the entire input column to the map function"

lilac/data/dataset.py
```diff
@@ -617,6 +618,8 @@ def map(
...
+        `num_jobs` must be 1 since a single process will be responsible for the entire input.
```
nsthorat, line 622: this isn't actually true -- the map worker will only see a slice, so you can use num_jobs > 1

nsthorat (approved):

lilac/data/dataset_map_test.py
```diff
@@ -949,3 +949,37 @@ def skip_first_and_last_letter(item: str) -> Item:
...
+  def text_len(items: Iterable[Item]) -> Iterable[Item]:
+    for item in items:
+      yield len(item['text'])
+
+  dataset = make_test_data([{'text': 'abcd'}, {'text': 'efghi'}])
+  with pytest.raises(Exception):
+    dataset.map(text_len, output_column='text_len', entire_input=True, num_jobs=2)
+
+
+def test_map_entire_input_with_input_path(make_test_data: TestDataMaker) -> None:
```
nsthorat, line 976: can you also add a unit test that makes sure we throw properly when the users didn't provide a 1:1 input to output, like we do for signals?

lilac/data/dataset.py
dsmilkov: Right, but we deny that combination of params for now because all the cases I need an iterable for require a global view of the world, and it's very easy to forget a num_jobs=-1 in there and get wrong results.

lilac/data/dataset.py
dsmilkov: done

lilac/data/dataset.py
dsmilkov: talked offline, here we are explicit that the map needs a global state

lilac/data/dataset_map_test.py
dsmilkov: done

lilac/data/dataset.py
```diff
@@ -617,6 +618,8 @@ def map(
       num_jobs: The number of jobs to shard the work, defaults to 1. When set to -1, the number of
         jobs will correspond to the number of processors.. If `num_jobs` is greater than the number
         of processors, it split the work into `num_jobs` and distribute amongst processors.
+      entire_input: If true, pass an iterable of the entire input column to the map function. When
```
brilee, line 621: This docstring isn't clear to me.

I think the diff here is that one is akin to

pd.DataFrame().apply(func, axis=0)

and the other is 
pd.Series().map(func)

dsmilkov: this is for the use case of doing something that needs to see the world (globally):
```py
def cluster(inputs: Iterable):
  global_state = do_stuff_on_entire_input(inputs)
  # Yield as many outputs as inputs, or we will yell
  for input in inputs:
      yield assign_cluster_id(input, global_state)
```

brilee: In that case I think it breaks the expectations/semantics of map enough that it ought to be an entirely separate method name. For example, .map() supports resumable based on per-item progress. What would that even look like if, say, the global state computation is interrupted?

dsmilkov: Nikhil had the same point, any ideas for the new name?

PR #899, opened 2023-11-29
Allow lilac to run tasks outside a running event loop.

nsthorat:
https://lilacai-nikhil-staging.hf.space/

### DIFF
--- a/lilac/tasks.py
+++ b/lilac/tasks.py
@@ -34,7 +34,6 @@ from distributed import Future, get_client, get_worker, wait
 from pydantic import BaseModel, TypeAdapter
 from tqdm import tqdm
 
-from .env import env
 from .utils import log, pretty_timedelta
 
 # nest-asyncio is used to patch asyncio to allow nested event loops. This is required when Lilac is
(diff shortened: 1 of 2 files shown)


lilac/tasks.py
```diff
@@ -121,10 +120,16 @@ def __init__(self, dask_client: Optional[Client] = None) -> None:
...
+    asynchronous = True
+    try:
+      asyncio.get_running_loop()
+    except RuntimeError as e:
+      asynchronous = False
+
     self.n_workers = multiprocessing.cpu_count()
     total_memory_gb = psutil.virtual_memory().total / (1024**3)
     self._dask_client = dask_client or Client(
-      asynchronous=not env('LILAC_TEST'),
```
dsmilkov, line 127: remind me why we disabled it during tests ? was it ever async during tests?

dsmilkov (approved):

lilac/tasks.py
nsthorat: We disabled it during tests for exactly this reason -- there is no running event loop!

So it was stupid I didn't check like this :)

brilee (approved):
image
this is how I feel about async python

PR #900, opened 2023-11-29
Refactor dataset/signal endpoints into separate module

brilee:
Just reducing the volume of code in the router_dataset.py module in anticipation of adding a bunch more dataset editing (column rename, dataset rename) endpoints.

### DIFF
--- a/lilac/server.py
+++ b/lilac/server.py
@@ -22,6 +22,7 @@ from . import (
   router_concept,
   router_data_loader,
   router_dataset,
+  router_dataset_signals,
   router_google_login,
   router_rag,
   router_signal,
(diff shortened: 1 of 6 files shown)


brilee (commented):
oops wrong thread.

lilac/server.py
```diff
@@ -113,6 +114,7 @@ def module_not_found_error(request: Request, exc: ModuleNotFoundError) -> JSONRe
 
 v1_router = APIRouter(route_class=RouteErrorHandler)
 v1_router.include_router(router_dataset.router, prefix='/datasets', tags=['datasets'])
+v1_router.include_router(router_dataset_signals.router, prefix='/datasets', tags=['datasets'])
```
nsthorat, line 117: do you need this since you import the router from the other file?

nsthorat (approved):

lilac/server.py
brilee: fixed

dsmilkov: can you rerun run_server_dev.sh? It will auto-gen a new typescript file now

dsmilkov (approved):
Can you run_server_dev , it should auto-gen another typesript file now that you moved stuff to

lilac/server.py
dsmilkov: also worth testing manually the UI

brilee: manually tested, and rerunning the gen didn't seem to change where the typescript file got generated?

brilee:
nikhil says it's based on the URL base name so that's why no new file is created

lilac/server.py
dsmilkov: ah, because you reused the same tag, which determines the file name, sgtm

PR #909, opened 2023-12-01
Small ergonomic fixes while writing the "code formatting" blog post

dsmilkov:
- Switch to single item by default instead of scroll
- Promote all long string fields (>=150 chars) to media fields
- Add `dataset.add_media_field` python API.
- Auto-promote any map outputs that are long to media fields.


lilac/data/dataset.py
```diff
@@ -387,6 +390,32 @@ def update_settings(self, settings: DatasetSettings) -> None:
     """Update the persistent settings for the dataset."""
     update_project_dataset_settings(self.namespace, self.dataset_name, settings, self.project_dir)
 
+  def add_media_field(self, path: Path, markdown: bool = False) -> None:
```
nsthorat, line 393: This is great -- I wonder if we can also just infer media fields by length whenever you run a map (augmenting the default media paths) so we dont have to do this explicitly at all in your post

nsthorat (approved):

lilac/data/dataset.py
dsmilkov: done

PR #912, opened 2023-12-02
Add blog post for diffing and `dataset.map`

dsmilkov:
https://lilac-386213--staging-s5a5qv9x.web.app/blog/better-curate.html

Also fix a few bugs along the way:
- add `HNSWLIB_NO_NATIVE` to fix HF deployment on AMD machines
- Fix tiny bug with sort dropdown not reflecting the auto-sort of semantic and concept search
- Fix a bug where boolean histograms were 1 and 0 instead of false and true
- Fix a bug where media paths can potentially point to columns we've manually deleted using `rm -r`
- Fix a bug in single item view when there are 0 results for a filter.
- Fix a bug where the user calls ll.set_project_dir() on directory that doesn't exist, and/or uses the "~" for homedir.
- Fix a bug with delayed requests with settings modal, when deployed on HF.
- Fix a bug where we forget the "expanded" state when fetching the next page in single item view
- Minimize re-rendering of items when scrolling or switching to the next example


docs/blog/better-curate.md
```diff
@@ -0,0 +1,153 @@
+# A better way to curate data
```
brilee, line 1: Using Lilac to curate a coding dataset

PR #920, opened 2023-12-05
Add threaded task execution.

nsthorat:
Adds a `type='threads'|'processes'` param to map() to determine whether to split jobs amongst threads or processes.

When a task type is threads, we create a ThreadPoolExecutor with the number of tasks submitted. This then spins up threads, and executes them.

To do this, I simulated the logging API of dask for threads.


lilac/data/dataset_map_test.py
```diff
@@ -71,15 +70,26 @@ def setup_teardown() -> Iterable[None]:
...
+  (-1, 'threads'),
+  (-1, 'processes'),
+  (1, 'threads'),
+  (1, 'processes'),
+  (2, 'threads'),
+  (2, 'processes'),
+]
+
+
+@pytest.mark.parametrize('params', JOBS_TYPE_PARAM)
```
dsmilkov, line 83: I've seen pytest encourage two params in this case, num_jobs, job_type, and it will itself do the 2x2 matrix of combinations

dsmilkov:
Nice work!

lilac/data/dataset_map_test.py
nsthorat: Done

lilac/data/dataset.py
```diff
@@ -645,8 +646,9 @@ def map(
         fields.
       resolve_span: Whether to resolve the spans into text before calling the map function.
       num_jobs: The number of jobs to shard the work, defaults to 1. When set to -1, the number of
-        jobs will correspond to the number of processors.. If `num_jobs` is greater than the number
+        jobs will correspond to the number of processors. If `num_jobs` is greater than the number
         of processors, it split the work into `num_jobs` and distribute amongst processors.
+      execution_type: The local execution type of the map. Either "threads" or "processes".
```
brilee, line 651: Add a sentence here about "Threads are better for network bound tasks, like calling a remote API; processes are better for compute bound tasks like running a local LLM."

lilac/data/dataset_duckdb.py
```diff
@@ -621,9 +629,10 @@ def _select_iterable_values(
     con = self.con.cursor()
 
     # Create a view for the work of the shard before anti-joining.
+    t_shard_table = f't_shard_{shard_id}'
```
brilee, line 670: oh, I noticed this name collision while I was going through this code. I'll probably fix this in a different way (subselects). this change is good for now.

lilac/data/dataset_duckdb.py
```diff
@@ -2614,7 +2625,11 @@ def map(
       jsonl_cache_filepaths.append(jsonl_cache_filepath)
 
     # Execute all the subtasks in parallel.
-    get_task_manager().execute_sharded(task_id, subtasks)
+    get_task_manager().execute_sharded(
```
brilee, line 2666: The prettier api here would be

get_task_manager(execution_type).executed_sharded(...)

lilac/data/dataset_duckdb.py
```diff
@@ -2633,10 +2648,15 @@ def map(
 
     map_field_root = map_schema.get_field(output_path)
 
+    map_source: str = ''
+    try:
+      map_source = inspect.getsource(map_fn)
```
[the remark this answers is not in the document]
brilee: not a security thing; the REPL is literally just running eval() on every line you enter, which leads to there being no associated source code file. The impl of getsource is approximately  `f.__module__.__file__.getlines(source_start_line, source_end_line)` so if there's no file there's no source code.

lilac/tasks.py
```diff
@@ -215,23 +220,32 @@ async def manifest(self) -> TaskManifest:
 
   def wait(self, task_ids: Optional[list[str]] = None) -> None:
```
brilee, line 221: This seems like two separate classes's method implementations interwoven into one... :(

brilee: I think the right design is to just have two implementations of TaskManagerInterface - each modality will ultimately have different weirdnesses and bugfixes and there's no need to try and preserve a common internal implementation.

lilac/data/dataset.py
nsthorat: Done

lilac/data/dataset_duckdb.py
nsthorat: Sweet

lilac/data/dataset_duckdb.py
nsthorat: agree, though changing out the task_manager is a large change since we assume the global singleton when we poll for tasks

lilac/data/dataset_duckdb.py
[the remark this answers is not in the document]
nsthorat: Ah, thanks

lilac/tasks.py
nsthorat: Agree, though just for the sake of getting this and the other things to solve in I'll probably file an issues since this class actually needs to be separated into two logically separate things -- one for managing Lilac tasks, and another for managing the type of executor, since the UI just uses this global singleton to get the state of pending tasks to render the UI (assumes everything is merged).

The differences then become a little smaller, and the task_manager would delegate to its children.

I filed an issue to track this: https://github.com/lilacai/lilac/issues/922

PR #925, opened 2023-12-05
Make sure the results of dataset.map() always returns an iterable.

nsthorat:


brilee (approved):

lilac/data/dataset_duckdb.py
```diff
@@ -913,19 +913,16 @@ def _reshard_cache(
...
-          {jsonl_cache_filepaths},
-          {f'columns={duckdb_schema(schema)},' if schema else ''}
-          hive_partitioning=false,
-          ignore_errors=true,
-          format='newline_delimited'
-        )
-      );
+    json_query = f"""
+      SELECT * FROM {'read_json' if schema else 'read_json_auto'}(
+        {jsonl_cache_filepaths},
```
dsmilkov, line 918: when do we cleanup this json_cache_filepath ? should we remove it after map finishes and we made a parquet file? In that case the json_query that the `result` iterable should execute is diff after map() completes.

dsmilkov (approved):

lilac/data/dataset_duckdb.py
nsthorat: Great question. We don't, we do it lazily. I am currently just returning the jsonl, if we change that this unit test will break and we can update it.

I think the design of that will coincide with how we think through restarts / limit / filters and resuming, so let's wait for that (actually, ideally, it just queries the dataset public API via select_iterable_values)

dsmilkov: Should we file a bug saying that succesful maps() write 2x the data (jsonl + parquet) and we dont cleanup. just to keep track of this.

PR #933, opened 2023-12-06
Add dataset.map support for limit/filter

brilee:
lots and lots of caveats to this code which I've documented.


lilac/data/dataset_duckdb.py
```diff
@@ -916,6 +923,14 @@ def _reshard_cache(
       schema = schema.model_copy(deep=True)
       schema.fields[ROWID] = Field(dtype=STRING)
 
+    # Potential bug: if a computation is interrupted, and then the num-workers, filtering, or limits
+    # are updated, then shard assignment may not be consistent. Then, two workers may have processed
```
nsthorat, line 927: the other thing that could happen -- what if num_rows changes and you dont have filters?

brilee: Hm.. we don't yet have any methods that can change the length of a dataset, but yeah, I see your point.

PR #947, opened 2023-12-08
Using `loky` (thin wrapper around `multiprocessing`) instead of dask

dsmilkov:
This change has a lot of benefits:
- Internal: dropping a large dependency and lots of complex code
- External: users can now define map in various places (notebooks, top-level, etc without the need to protect surrounding score with `if __name__ == __main__` blocks, and avoiding various pickle serialization errors.


lilac/tasks.py
```diff
@@ -724,3 +730,17 @@ def set_worker_task_progress(
...
 
   set_worker_steps(task_id, steps)
+
+
+def check_worker_tries_to_write_to_db(func_name: str) -> None:
+  """Raises an error when a worker tries to write to the database."""
+  if get_is_dask_worker():
+    raise RuntimeError(
+      f"""
+      Your `map` function or its surrounding code is calling {func_name}(...).
```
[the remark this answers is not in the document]
brilee: see textwrap.dedent()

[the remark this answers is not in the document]
dsmilkov: outdated

dsmilkov:
hey guys, please re-review, it's a completely different PR at this point

lilac/conftest.py
```diff
@@ -30,7 +30,7 @@ def _make_test_data(items: list[Item], schema: Optional[Schema] = None) -> Datas
 
 
 @pytest.fixture(scope='function')
-def test_dask_logger(tmp_path: pathlib.Path) -> Generator:
+def test_jsonl_logger(tmp_path: pathlib.Path) -> Generator:
```
nsthorat, line 33: jsonl logger is sort of an impl detail, how about test_process_logger

lilac/data/dataset_duckdb.py
```diff
@@ -2759,12 +2726,7 @@ def map(
       type=execution_type,
       subtasks=subtasks,
     )
-    show_progress(
-      task_step_id=(task_id, 0), total_len=manifest.num_items, description=progress_description
-    )
-
-    # Wait for the tasks to finish before reading the outputs.
-    get_task_manager().wait([task_id])
+    block_and_show_progress(task_id, description=progress_description)
```
nsthorat, line 2729: nit: show_progress_and_block

lilac/load.py
```diff
@@ -39,8 +33,7 @@ def load(
   project_dir: Optional[Union[str, pathlib.Path]] = None,
   config: Optional[Union[str, pathlib.Path, Config]] = None,
   overwrite: bool = False,
-  task_manager: Optional[TaskManager] = None,
-  load_task_id: Optional[str] = None,
+  use_threads: bool = False,
```
nsthorat, line 36: how about using the `execution_type` we use in other places?

lilac/load_test.py
```diff
@@ -305,35 +291,31 @@ def test_load_twice_no_overwrite(
...
-  assert compute_signal_mock.call_count == 1
-  assert compute_embedding_mock.call_count == 1
+  # assert test_jsonl_logger.get_logs() == ['compute embedding called', 'compute signal called']
 
   second_manifest = get_dataset('namespace', 'test').manifest()
-
   assert first_manifest == second_manifest
 
+  assert (
+    'Signal  TestSignal({"signal_name":"test_signal"}) already exists' in capsys.readouterr().out
```
nsthorat, line 309: nit: double space here

lilac/server.py
```diff
@@ -57,26 +56,11 @@ def custom_generate_unique_id(route: APIRoute) -> str:
...
   """Context manager for the lifespan of the application."""
   if env('LILAC_LOAD_ON_START_SERVER', False):
-    task_manager = get_task_manager()
-    task_id = task_manager.task_id(
-      'Loading from project config... ',
-      description='This can be disabled by setting the environment variable '
-      'LILAC_LOAD_ON_START_SERVER=false',
-    )
-    task_manager.execute(task_id, 'processes', _load, task_id)
+    load(project_dir=get_project_dir(), overwrite=False)
```
nsthorat, line 63: might be worth explicitly setting a type here

lilac/sources/parquet_source.py
```diff
@@ -131,9 +131,9 @@ def source_schema(self) -> SourceSchema:
 
   @override
   def load_to_parquet(
-    self, output_dir: str, task_step_id: Optional[TaskStepId] = None
+    self, output_dir: str, task_shard_id: Optional[TaskShardId] = None
```
nsthorat, line 134: funky api here -- this would never take a shard index, so mayb e best to just have it as task_id?

lilac/tasks.py
```diff
@@ -70,47 +64,37 @@ class TaskType(str, Enum):
...
-  shard_progresses: list[tuple[int, tuple[int, int]]] = []
+  current_index: int
+  estimated_len: Optional[int]
 
 
 TaskExecutionType = Literal['processes', 'threads']
 
 
-class TaskInfo(BaseModel):
+@dataclass
```
nsthorat, line 78: not a pydantic model?

nsthorat (approved):
Nice work, this is crazy!

lilac/tasks.py
```diff
@@ -562,165 +390,29 @@ def show_progress(
...
-  start_time = time.time()
-  # Reduce the emit frequency if there are multiple shards to reduce the size of the event stream.
-  emit_every_sec = EMIT_EVERY_SEC if not shard_count else EMIT_EVERY_SEC * shard_count
-  # Add jitter to the emit frequency to avoid all workers emitting at the same time.
-  jitter_sec = random.uniform(0, emit_every_sec)
-  last_emit = time.time() - EMIT_EVERY_SEC - jitter_sec
+  it_idx = initial_index if initial_index else 0
+  shard_info = TaskShardInfo(current_index=it_idx, estimated_len=estimated_len)
+  emit_every_sec = EMIT_EVERY_SEC
+  # Add +/- 20% jitter to the emit frequency to avoid all workers emitting at the same time.
```
brilee, line 405: oh, I was the one who suggested this to nikhil and this edit breaks the thing I was going for.

Basically if you have 8 workers with 8 second reporting time, then it would suck to have zero task progress update for  8 seconds, and then a bunch of updates at once. Ideally you would get a trickle of updates once a second from each worker.

Anyway I'm now not even sure the math would work out correctly because if worker 1 says "I did 100 items" at T=1, worker 2 says "I did 200 items" at T=2, then the computed progress would be "300 items @ 2 sec", whereas the true answer would have been 1600 items at T=2 from all 8 workers.

brilee (commented):
perhaps let's step through in person tomorrow?

lilac/conftest.py
dsmilkov: done

lilac/data/dataset_duckdb.py
dsmilkov: done

lilac/load.py
dsmilkov: done

lilac/load_test.py
dsmilkov: done

lilac/server.py
dsmilkov: done

lilac/sources/parquet_source.py
dsmilkov: done

lilac/tasks.py
dsmilkov: yeah, this objects gets pickled a bunch and sent over sockets, and dataclass is much lighter for that (pydantic is better for json/dict serializability)

lilac/tasks.py
dsmilkov: yeah, let's revisit this tomorrow -- i can change it to uniform(0, 1) but generally with the new approach map doesn't slow down with number of workers, unlike the previous approach

brilee (approved):

lilac/tasks.py
dsmilkov: reverted to adjust for number of workers

PR #951, opened 2023-12-08
Add special support for a DELETED label

brilee:


lilac/data/dataset.py
```diff
@@ -590,6 +609,30 @@ def remove_labels(
...
+    searches: Optional[Sequence[Search]] = None,
+    filters: Optional[Sequence[FilterLike]] = None,
+  ) -> int:
+    """Deletes rows from the dataset.
+
+    Returns the number of deleted rows.
+    """
+    return self.add_labels(DELETED_LABEL_NAME, row_ids, searches, filters)
+
+  def undelete_rows(
```
nsthorat, line 624: maybe restore? not a strong opinion

lilac/data/dataset_export_test.py
```diff
@@ -168,27 +169,37 @@ def test_label_and_export_by_excluding(
   make_test_data: TestDataMaker, tmp_path: pathlib.Path
 ) -> None:
   dataset = make_test_data([{'text': 'a'}, {'text': 'b'}, {'text': 'c'}])
-  dataset.add_labels('delete', ['2', '3'])
+  dataset.add_labels(DELETED_LABEL_NAME, ['2', '3'])
```
nsthorat, line 172: maybe separate into two tests, one for a custom label name, and one that calls the public API for delete

lilac/data/dataset_map_test.py
```diff
@@ -602,6 +602,48 @@ def _map_fn(items: list[Item]) -> list[Item]:
   ]
 
 
+@pytest.mark.parametrize('num_jobs', [-1, 1, 2])
+@pytest.mark.parametrize('execution_type', TEST_EXECUTION_TYPES)
+def test_map_include_deleted(
```
nsthorat, line 607: maybe a test for exclude deleted?

lilac/data/dataset_select_groups_test.py
```diff
@@ -37,6 +37,23 @@ def test_flat_data(make_test_data: TestDataMaker) -> None:
...
+    {'name': 'Name1', 'age': 34, 'active': False},
+    {'name': 'Name2', 'age': 45, 'active': True},
+    {'age': 17, 'active': True},  # Missing "name".
+  ]
+  dataset = make_test_data(items)
+
+  result = dataset.select_groups(leaf_path='name')
+
+  assert result.counts == [('Name1', 1), ('Name2', 1), (None, 1)]
+  dataset.add_labels(DELETED_LABEL_NAME, filters=[('name', 'equals', 'Name1')])
```
nsthorat, line 51: call public api?

nsthorat (approved):
you are on a roll

High-level question:

- What about dataset.manifest().num_rows? This maybe should reflect the count?
- Should we expose a num_deleted_rows?

lilac/data/dataset_duckdb.py
```diff
@@ -2049,6 +2072,28 @@ def select_rows_schema(
...
+      sqlite_cur.execute(
+        f"""
+        CREATE TABLE IF NOT EXISTS "{label_name}" (
+          {ROWID} VARCHAR NOT NULL PRIMARY KEY,
+          label VARCHAR NOT NULL,
+          created DATETIME)
+      """
+      )
+    return labels_filepath
+
```
dsmilkov, line 2096: just to confirm here, will `add_labels('some_new_label', filters=..., searches=...` ignore labeling all previously deleted rows?

lilac/data/dataset.py
```diff
@@ -590,6 +609,30 @@ def remove_labels(
...
+    filters: Optional[Sequence[FilterLike]] = None,
+  ) -> int:
+    """Undeletes rows from the dataset.
+
+    Returns the number of restored rows.
+    """
+    return self.remove_labels(DELETED_LABEL_NAME, row_ids, searches, filters)
+
   @abc.abstractmethod
   def stats(self, leaf_path: Path) -> StatsResult:
```
dsmilkov, line 637: the duckdb implementation of `stats` should ignore any deleted rows, probably we should also remove the cache decorator in duckdb.stats, or make it smarter by invalidating only when a row is deleted.

dsmilkov (approved):
This is amazing! 

Good to add coverage for deleting rows and then calling compute_signal , compute_embeddings, stats, add_labels, remove_labels. Can be a future PR.

lilac/data/dataset_export_test.py
brilee: I replaced with the public API. Rethought some of the semantics around export; I think the export should treat everything in a transparent way. So the UI should show the __deleted__ label as just another label that is by default checked in the "exclude" list.

brilee:
Frontend implementation notes:

- need to add include_excluded to HTTP endpoints.
- when a row/rows are deleted, must invalidate a bunch of network requests so that the UI refreshes.
- need row-level delete and "delete all" buttons.
- export data window will show the `__deleted__` label; this label should be checked by default.
- there should be a global toggle for "show deleted items" that propagates through all API calls.
- when "show deleted items" is toggled, a button for restoring a deleted row should appear.
-

lilac/data/dataset_select_groups_test.py
brilee: done

lilac/data/dataset_map_test.py
brilee: this test shows examples of both include/exclude deleted. Will update name to clarify test intent.

PR #952, opened 2023-12-09
Switch to monaco for the main viewer.

nsthorat:
High-level: switch over to monaco for the main viewer. This gives us a lot of control with much simpler code.

https://github.com/lilacai/lilac/assets/1100749/8a254d97-cf96-4eeb-9f75-30172c3b4539

Other changes:
- Make the markdown renderer a toggle now. The search is completely broken in the markdown renderer, so this is now just a toggle to switch between monaco and the svelte markdown viewer (with no searches).
image

- Allow highlighting text, and searching by keyword, semantic search, or adding to a concept (only if the concept is an active search).
image

Details:
- Introduce <ItemMediaTextContent> which uses monaco for the main text viewer.
- When spans are highlighted, we do a linear pass to find them, and build the span hover card.

Notes:
- We still have StringSpanHighlight for now. This is used in the concept preview, insight items, and the signal view. We can clean this up after.

Demo:
https://lilacai-nikhil-staging.hf.space/datasets#local/OpenOrca-10k
https://lilacai-nikhil-staging.hf.space/datasets#local/OpenHermes-2.5-100k

NOTE: I tried very hard to remove "Command Palette" from the right click menu.. it is extremely brittle and hard to do.

NOTE: tests are failing because it relies on a previous PR.


dsmilkov:
This is bonkers cool!

Probably best to walk me through the code on Monday. Sharing here some thoughts/questions:

- i had asked recently the cohere team if they prefer for lilac to switch to monospace font by default, and their answer was that monospace can be a toggle, but that the default font we use is easier to read text and understand content faster. Can we make monaco render with a regular font by default and a toggle/setting for monospace?
- Is it possible to auto-scroll monaco to the first line it contains a certain keyword/highlighted region, or collapse the regions around the highlighted text, just like we do in string span highlight?
- not a priority: Any ideas on how to make highlighting work in markdown mode?

dsmilkov (approved):

nsthorat:
Thanks!

- I pinged them to see if this is okay. I purposefully made the markdown preview a toggle so that should work for them (it's exactly how VSCode would do it). I also chose a monospace font that is "pretty" so it's readable in that view.
- Auto-scroll and line-linking is the next PR.
- I don't think we should do it at all. You would have to reach into the markdown renderer, and it's just not worth the complexity (and it's a lot of complexity). I think we should think of it as VSCode -- they never do anything on top of a markdown preview, it's always on the original, and it's just a preview.

I'll get this in just so I can keep going -- I'm gonna allow you to deep-link to a selection, or a line in the follow up.

PR #957, opened 2023-12-12
Fix bug with loading dataset and settings.

nsthorat:

Broken in https://github.com/lilacai/lilac/pull/912

### DIFF
--- a/web/blueprint/src/lib/components/datasetView/DatasetLoadingPage.svelte
+++ b/web/blueprint/src/lib/components/datasetView/DatasetLoadingPage.svelte
@@ -6,6 +6,7 @@
     queryDatasetSchema,
     querySelectRows,
     querySelectRowsSchema,
+    querySettings,
     updateDatasetSettingsMutation
   } from '$lib/queries/datasetQueries';
   import {queryTaskManifest} from '$lib/queries/taskQueries';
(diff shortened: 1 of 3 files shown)


web/blueprint/src/lib/components/datasetView/DatasetLoadingPage.svelte
```diff
@@ -33,6 +34,14 @@
   const datasetViewStore = createDatasetViewStore(namespace, datasetName);
   setDatasetViewContext(datasetViewStore);
 
+  $: settingsQuery = loadTaskComplete ? querySettings(namespace, datasetName) : null;
+  $: newSettings =
```
dsmilkov, line 38: where is newSettings being used?

dsmilkov (approved):

dsmilkov:
What's the actual bug / repro? Curious since we double bind settings from an inner component

web/blueprint/src/lib/components/datasetView/DatasetLoadingPage.svelte
nsthorat: its what is bound to the component below. was set to null because previously the settings component would fetch it.

dsmilkov (approved):

PR #959, opened 2023-12-12
Eliminate setup count call from parquet_source.

brilee:
The count() call would access all N shards even if pseudosampling was on. I defer the count so that it only accesses the final written file.


nsthorat (approved):
l o l

PR #961, opened 2023-12-13
Reduce import times

brilee:
Major change is to move umap and spacy to lazy import - they're 70% of the import time together. So imports are down from 4.5 -> 1.5 seconds

I cleaned up a few things here and there, but the remaining import time is due to:

-40% numpy ecosystem (numpy, sklearn, scipy, pandas)
-20-30% fastapi (could be optimized by splitting off lilac the server from lilac the library
-20% compiling pydantic models
-10-20% long tail of imports

### DIFF
--- a/lilac/splitters/spacy_splitter.py
+++ b/lilac/splitters/spacy_splitter.py
@@ -33,9 +33,9 @@ def embed_tokenizer_BoW(texts: list[str]) -> np.ndarray:
 
 @functools.cache
 def get_spacy() -> 'spacy.Language':
+  """Lazily instantiate and return a singeton SpaCy sentencizer object."""
   import spacy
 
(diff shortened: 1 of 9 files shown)


lilac/splitters/spacy_splitter.py
```diff
@@ -30,7 +32,9 @@ def embed_tokenizer_BoW(texts: list[str]) -> np.ndarray:
 
 
 @functools.cache
-def get_spacy() -> spacy.Language:
+def get_spacy() -> 'spacy.Language':
+  import spacy
```
nsthorat, line 37: some comment here?

nsthorat (commented):
wooooo

dsmilkov (approved):

PR #962, opened 2023-12-13
Add sugar for `ll.from_huggingface()`

dsmilkov:
```python
import lilac as ll
from datasets import load_dataset
hf_ds = load_dataset('Open-Orca/SlimOrca-Dedup')
ds = ll.from_huggingface(hf_ds)
# The dataset is namespace=local, name=SlimOrca-Dedup
```

### DIFF
--- a/lilac/load_dataset.py
+++ b/lilac/load_dataset.py
@@ -10,11 +10,12 @@ poetry run python -m lilac.load_dataset \
 import os
 import pathlib
 import uuid
-from typing import Iterable, Optional, Union
+from typing import Iterable, Optional, Union, cast
 
 import pandas as pd
 from datasets import Dataset as HFDataset
-from datasets import DatasetDict, load_dataset
+from datasets import DatasetDict as HFDatasetDict
+from datasets import load_dataset
 
 from .config import DatasetConfig
 from .data.dataset import Dataset, default_settings
@@ -69,7 +70,7 @@ def from_dicts(
 
 
 def from_huggingface(
-  dataset: Union[str, HFDataset, DatasetDict],
+  dataset: Union[str, HFDataset, HFDatasetDict],
   namespace: str = 'local',
   name: Optional[str] = None,
   overwrite: bool = False,
@@ -82,23 +83,19 @@ def from_huggingface(
     name: The Lilac name of the dataset to create. Defaults to the name of the HuggingFace dataset.
     overwrite: Whether to overwrite the dataset if it already exists.
   """
-  dataset_dict: DatasetDict
   if isinstance(dataset, str):
-    dataset_dict = load_dataset(dataset)
-  elif isinstance(dataset, HFDataset):
-    dataset_dict = DatasetDict({'train': dataset})
-  else:
-    dataset_dict = dataset
+    dataset = cast(Union[HFDataset, HFDatasetDict], load_dataset(dataset))
 
   if not namespace:
     namespace = 'local'
 
   if not name:
     # Infer name from the dataset instance.
-    if dataset_dict.keys():
-      ds = list(dataset_dict.values())[0]
-      if ds.info.dataset_name:
-        name = ds.info.dataset_name
+    split: HFDataset = dataset
+    if isinstance(dataset, HFDatasetDict):
+      split = list(cast(HFDatasetDict, dataset).values())[0]
+    if split.info.dataset_name:
+      name = split.info.dataset_name
 
   if not name:
     raise ValueError('`name` must be defined since it could not be inferred from the dataset.')
@@ -106,7 +103,7 @@ def from_huggingface(
   config = DatasetConfig(
     namespace=namespace,
     name=name,
-    source=HuggingFaceSource(dataset_name=name, dataset_dict=dataset_dict),
+    source=HuggingFaceSource(dataset_name=name, dataset=dataset),
   )
   return create_dataset(config, overwrite=overwrite)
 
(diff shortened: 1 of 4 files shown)


lilac/load_dataset.py
```diff
@@ -65,6 +66,58 @@ def from_dicts(
   return create_dataset(config, overwrite=overwrite)
 
 
+def from_hf(
```
brilee, line 69: I'd write it out in full: `def from_huggingface`

lilac/load_dataset.py
```diff
@@ -65,6 +66,58 @@ def from_dicts(
...
+  dataset_name: str,
+  namespace: Optional[str] = None,
+  name: Optional[str] = None,
+  overwrite: bool = False,
```
brilee, line 73: There's no need to bloat the api surface area complexity of this sugar; I would strip most of these kwargs. If you need the full functionality then just call ll.load_dataset(ll.DatasetConfig(...))

brilee: The most common use case I think would be to point it at a live HF dataset on the web? Second most common would be one that's already in memory as a hf.Dataset instance.

lilac/load_dataset.py
```diff
@@ -65,6 +66,58 @@ def from_dicts(
...
+    config_name: The name of the dataset configuration to load from HuggingFace.
+    split: The split to load from the dataset.
+    sample_size: The number of samples to load from the dataset.
+    token: The HuggingFace API token to use.
+    revision: The revision of the dataset to load.
+    load_from_disk: Whether to load the dataset from disk.
+  """
+  if '/' in dataset_name:
+    parts = dataset_name.split('/')
+    hf_namespace, hf_name = parts[-2], parts[-1]
```
brilee, line 98: reusing the hf namespace as a local namespace seems potentially weird; you'd end up with a bunch of single-use namespaces on the sidebar. I'd retain the full dataset_name as the dataset name and ask user to explicitly provide a local namespace name

lilac/load_dataset.py
nsthorat: I sort of think that the in-memory one might be the most important code path.

Users already know the HF API, we don't have to mirror all of their params, and we can just take a HF DS instance and load it (avoiding propagating tokens, from_disk, and all that).

e.g.

```
import lilac as ll
from datasets import load_dataset

hf_ds = load_dataset("lhoestq/demo1")
ll_ds = ll.from_huggingface(hf_ds, 'local', 'demo1')
```

OR 


```
import lilac as ll
from datasets import load_dataset

hf_ds = load_dataset("lhoestq/demo1")
hf_ds = hf_ds.map(....)

# Post huggingface map is very easy now
ll_ds = ll.from_huggingface(hf_ds, 'local', 'demo1')
```

lilac/load_dataset.py
nsthorat: +1, can default namespace to local

lilac/load_dataset.py
dsmilkov: done

lilac/load_dataset.py
dsmilkov: done, made it go through a hf dataset instance, OR string for the convenience in loading directly from hub, but removed all extra params. if the user wants to specify params, they need to provide a hf_dataset instance.

lilac/load_dataset.py
dsmilkov: local is currently underused , and the org name of the dataset is a good context signal , but i reverted to 'local' for now. we can always make it smarter later.

lilac/load_dataset.py
```diff
@@ -65,6 +68,50 @@ def from_dicts(
   return create_dataset(config, overwrite=overwrite)
 
 
+def from_huggingface(
+  dataset: Union[str, HFDataset, DatasetDict],
+  namespace: Optional[str] = None,
```
nsthorat, line 73: namespace='local' here for documentation

lilac/load_dataset.py
```diff
@@ -65,6 +68,50 @@ def from_dicts(
...
+  name: Optional[str] = None,
+  overwrite: bool = False,
+) -> Dataset:
+  """Load a dataset from HuggingFace.
+
+  Args:
+    dataset: A HF dataset or the name of the dataset registered in HF hub.
+    namespace: The Lilac namespace for the loaded dataset. Defaults to `local`.
+    name: The Lilac name of the dataset to create. If not defined, defaults to the name of the HF
```
nsthorat, line 82: shorten, remove "If not defined,"

nsthorat, line 82: HF => HuggingFace

lilac/sources/huggingface_source.py
```diff
@@ -143,6 +144,9 @@ class HuggingFaceSource(Source):
   name: ClassVar[str] = 'huggingface'
 
   dataset_name: str = PydanticField(description='Either in the format `user/dataset` or `dataset`.')
+  dataset_dict: Optional[DatasetDict] = PydanticField(
+    description='The pre-loaded HF dataset dict', exclude=True, default=None
```
nsthorat, line 148: maybe huggingface_dataset instead? it's not obvious from their docs this is a DatasetDict

and then more in the description here -- "A pre-loaded HuggingFace dataset object. For instance, ds=load_dataset('imdb')"

nsthorat (approved):

lilac/sources/huggingface_source.py
dsmilkov: since this is still public API, i relaxed the dataset to be either HF DatasetDict or HF Dataset, and we deal with normalizing here instead of the sugar method.

PR #966, opened 2023-12-14
Add Jina V2 embeddings

dsmilkov:
The current implementation trims the document and doesn't do any splits. It leads to a much simpler way to browser similar things and I need it at doc level to find prompts that are unrelated to the answer (bad data)

Also:
 - Add additional schema validation
 - Move test only schema inference code to a test only util
 - Fix an issue with shutting down the process pool

### DIFF
--- a/lilac/data/dataset_test_utils.py
+++ b/lilac/data/dataset_test_utils.py
@@ -3,7 +3,8 @@ import json
 import os
 import pathlib
 from copy import deepcopy
-from typing import Any, ClassVar, Optional, Type
+from datetime import datetime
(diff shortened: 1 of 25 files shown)


lilac/data/dataset_test_utils.py
```diff
@@ -44,14 +56,68 @@ class TestSource(Source):
   name: ClassVar[str] = 'test_source'
 
 
+def _infer_dtype(value: Item) -> DataType:
```
dsmilkov, line 59: these are test only utils that i moved away from schema.py

brilee (approved):

PR #974, opened 2023-12-15
Fix memory leak caused by Iterable/Iterator mixups

brilee:
Major change here is to strictly enforce that compute(), map(), and related functionality returns an Iterator, not an Iterable, since the former is more likely to be O(1) memory (misuses of tee aside...), whereas the latter is more likely to be O(N) memory.

Still haven't figured out the duckdb memory leak; seems like something in our mess of join views and nested views is causing duckdb to pull the entire parquet file into memory, even if the final query is only requesting row_id and one other column. The ORDER BY ROWID is likely contributing to the issue, but removing the ORDER BY does not fully fix the "pull it all into memory" issue.

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -92,6 +92,7 @@ from ..signals.filter_mask import FilterMaskSignal
 from ..signals.semantic_similarity import SemanticSimilaritySignal
 from ..signals.substring_search import SubstringSignal
 from ..source import NoSource, SourceManifest
+from ..sources.duckdb_utils import duckdb_setup
 from ..tasks import (
   TaskExecutionType,
   TaskFn,
@@ -304,6 +305,11 @@ class DuckDBQueryParams(BaseModel):
   include_deleted: bool = False
 
 
+def consume_iterator(iterator: Iterator[Any]) -> None:
+  for _ in iterator:
+    pass
+
+
 class DatasetDuckDB(Dataset):
   """The DuckDB implementation of the dataset database."""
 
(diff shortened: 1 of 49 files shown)


lilac/data/dataset_duckdb.py
```diff
@@ -304,6 +304,12 @@ class DuckDBQueryParams(BaseModel):
   include_deleted: bool = False
 
 
+def _consume_iterator(iterator: Iterator[Any]) -> None:
```
nsthorat, line 307: https://stackoverflow.com/questions/50937966/fastest-most-pythonic-way-to-consume-an-iterator

I looked this up recently and apparently collections.deque is the most efficient way to do this

nsthorat (approved):
SHIP, amazing!

lilac/data/dataset_duckdb.py
brilee: lol, the overhead reduction is small enough that I'd prefer the more obvious implementation :P

lilac/data/dataset_utils.py
```diff
@@ -305,21 +304,18 @@ def flatten_keys(
...
-    for value in x:
-      if value is not None:
-        yield value
-
+  sparse_input = iter(sparse_input)
   sparse_input, sparse_input_2 = itertools.tee(sparse_input, 2)
-  dense_input = densify(sparse_input_2)
-  dense_output = iter(func(dense_input))
+  dense_input = cast(Iterator[Tin], filter(lambda x: x is not None, sparse_input_2))
+  dense_output = func(dense_input)
```
dsmilkov, line 313: the iter() wrapper here was added at some point because custom signal implementations (outside lilac) might return a list from compute() at which point you get a very strange error where next(...) on line 318 fails. 

Two options:
 - a) we keep wrapping in iter(...) so we are relaxed about signal implementations 
 - b) we yell much earlier that the signal didn't return an Iterator with a nice stack trace / error msg

PR #981, opened 2023-12-18
Add `dataset.cluster()`

dsmilkov:
- Add `dataset.cluster()`, no UI yet
- Fix a bug with signal menu in schema
- Fix a bug with pagination of groups
- Make the group by header visually aligned with the row header
- Add `Signal.output_type` which can be embedding or a cluster


lilac/data/dataset.py
```diff
@@ -454,6 +455,11 @@ def compute_signal(
     """
     pass
 
+  def cluster(self, embedding: str, path: Path) -> None:
```
nsthorat, line 458: maybe path first

and maybe we should add a type here with embedding optional? minhash is pretty popular

web/blueprint/src/lib/components/contextMenu/SchemaFieldMenu.svelte
```diff
@@ -79,7 +81,7 @@
           })}
       />
     {/if}
-    {#if !isSignal}
+    {#if isComputable}
```
nsthorat, line 84: maybe canComputeSignal?

nsthorat (approved):

lilac/data/dataset.py
dsmilkov: done

web/blueprint/src/lib/components/contextMenu/SchemaFieldMenu.svelte
dsmilkov: done

PR #987, opened 2023-12-19
Fix some small UI bugs.

nsthorat:
- We showed null randomly
- Fix some issues with loading jumping, show progress bars before fonts are ready.
- Fix issue with vertical alignment of loading on the row id.

### DIFF
--- a/web/blueprint/src/lib/components/datasetView/ItemMedia.svelte
+++ b/web/blueprint/src/lib/components/datasetView/ItemMedia.svelte
@@ -51,6 +51,7 @@
   export let rootPath: Path | undefined = undefined;
   // The showPath is a subset of the path that will be displayed for this node.
   export let showPath: Path | undefined = undefined;
+  export let isFetching: boolean | undefined = undefined;
 
   // Choose the root path which is up to the point of the next repeated value.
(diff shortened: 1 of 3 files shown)


web/blueprint/src/lib/components/datasetView/ItemMedia.svelte
```diff
@@ -284,9 +285,9 @@
       {/if}
 
       <div class="grow pt-1">
-        {#if row == null}
+        {#if isFetching}
```
dsmilkov, line 288: i think row == null is equivalent to isFetching since if you look at the parent (RowItem), row is null until we get all the data.

If we passed a sub-set of row via `<svelte:self row={row[some_property]}`, then it would be different, but it looks like we are always passing down the top level row

dsmilkov (approved):

web/blueprint/src/lib/components/datasetView/ItemMedia.svelte
nsthorat: It's slightly different, this fixed the bug.

Cant pass subsets of row, would be a massive, not-very-valuable change (I think we've chatted about that before).

PR #993, opened 2023-12-21
Add topic clustering in `dataset.cluster()`

dsmilkov:
Currently we call a signal, then call a `dataset.transform` to name the clusters. They will also become internal detail in a future PR.

Also
- fix a bug with deep_unflatten and rename to array_unflatten
- fix a bug with select_iterable_values when we have input_path and combine=True
- fix a bug with progress reporting not finishing with 100% at the end of map
- fix a bug with signal disposal
- fix a bug with hdbscan when the dataset size is small (e.g. smaller than min_cluster size)

### DIFF
--- a/lilac/data/clustering.py
+++ b/lilac/data/clustering.py
@@ -46,7 +46,9 @@ def summarize_instructions(ranked_docs: list[tuple[str, float]]) -> str:
     f'INSTRUCTION {i+1}\n{_shorten(doc)}\nEND_INSTRUCTION {i+1}' for i, doc in enumerate(docs)
   ]
   input = '\n'.join(texts)
+  print(input)
   title = _openai_client().chat.completions.create(
+    # model='gpt-4-1106-preview',
     model='gpt-3.5-turbo',
     response_model=Title,
     temperature=0.0,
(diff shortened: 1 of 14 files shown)


lilac/data/dataset.py
```diff
@@ -455,12 +446,88 @@ def compute_signal(
...
+      path: The path to the text field to cluster.
+      embedding: The pre-computed embedding to use.
+      output_column: The name of the output column to write to. When `nest_under` is False
+        (the default), this will be the name of the top-level column. When `nest_under` is True,
+        the output_column will be the name of the column under the path given by `nest_under`.
+      nest_under: The path to nest the output under. Defaults to the input `path`, so it gets
+        hierarchically shown in the UI.
+      min_cluster_size: The minimum number of docs in a cluster.
+      topic_fn: A function that takes a list of (doc, membership_score) tuples and returns a
+        single topic. This is used to compute the topic for a given cluster of docs. It defaults
```
nsthorat, line 471: topic title

lilac/data/clustering.py
```diff
@@ -0,0 +1,67 @@
...
+  except ImportError:
+    raise ImportError(
+      'Could not import the "openai" python package. '
+      'Please install it with `pip install openai`.'
+    )
+
+  return instructor.patch(openai.OpenAI())
+
+
+def _shorten(text: str) -> str:
```
nsthorat, line 27: _snippet_to_prefix_suffix or something

nsthorat (approved):

lilac/data/clustering.py
dsmilkov: done

PR #994, opened 2023-12-21
Fix map

brilee:
There is now a central manager that computes rows remaining, and then sends those rows over IPC to the process/threadpool, instead of each worker independently trying to compute its shard of remaining work.

This means that each process no longer needs to make its own duckdb query, so that memory usage of duckdb is limited to just the manager process. It also removes all sorts of continuation bugs where changing the filter/limit settings might cause shard reassignment and shards unable to find their own previous work.

Joblib doesn't have an easy way to do job ids (see https://github.com/joblib/joblib/issues/1008) so I have dropped the job id functionality. 

I can now run a 10 million row map (peak memory consumption 24GB, still all due to DuckDB) with 10 processes. At HEAD we would freeze on a 1million row map with 10 processes.

Task progress reporting is vastly simplified since the manager bottlenecks and receives the finished work as a reply from each worker, allowing a trivial wrap in tqdm.


lilac/tasks.py
```diff
@@ -1,47 +1,25 @@
...
 
 from .env import env
-from .utils import log, pretty_timedelta
-
-# nest-asyncio is used to patch asyncio to allow nested event loops. This is required when Lilac is
-# run from a Jupyter notebook.
-# https://stackoverflow.com/questions/46827007/runtimeerror-this-event-loop-is-already-running-in-python
-if hasattr(builtins, '__IPYTHON__'):
-  # Check if in an iPython environment, then apply nest_asyncio.
-  nest_asyncio.apply()
```
nsthorat, line 44: just verify that this still works in jupyter?

if so, you can remove it from pyproject too

lilac/tasks.py
```diff
@@ -200,225 +111,81 @@ def task_id(
...
-  return TASK_SHARD_PROXY[task_shard_id]
 
 
 def get_task_manager() -> TaskManager:
   """The global singleton for the task manager."""
   global _TASK_MANAGER
-  if _TASK_MANAGER:
-    return _TASK_MANAGER
-  _TASK_MANAGER = TaskManager()
+  if not _TASK_MANAGER:
```
nsthorat, line 151: you may be able to use @functools.cache here now, I did this so it could be overwritten in a unit tests with dask

lilac/tasks_test.py
```diff
@@ -1,18 +0,0 @@
-"""Tests for tasks.py."""
-
-
-from .tasks import TaskManager
-
-
-def test_task_manager_outside_event_loop() -> None:
```
nsthorat, line 7: worth keeping these tests?

nsthorat (approved):

lilac/tasks.py
brilee: done

lilac/tasks_test.py
brilee: there's no complicated proxy stuff happening in the __init__ and stop() doesn't exist anymore... so probably not?

PR #997, opened 2023-12-21
Use cuml for clustering when possible

dsmilkov:


lilac/embeddings/gte.py
```diff
@@ -54,7 +55,16 @@ def compute(self, docs: Iterable[RichData]) -> Iterator[Item]:
 
   @override
   def teardown(self) -> None:
+    self._model.cpu()
```
nsthorat, line 58: should this be a util too since it's non obvious and copy pasted?

nsthorat (approved):
wowza

PR #999, opened 2023-12-22
`dataset.map` can now nest_under any repeated

dsmilkov:
You can call:

```python
before_row == {
  people: [
    {name: 'A', address: '...'},
    {'name': 'B', address: '...'}
  ]
}

dataset.map(geo_locate, input_path='people.*.address', nest_under='people.*', output_column='geo')

after_row == {
  people: [
    {name: 'A', address: '...', geo: '...'},
    {'name': 'B', address: '...', geo: '...'}
  ]
}
```

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -2652,7 +2652,6 @@ class DatasetDuckDB(Dataset):
       if not manifest.data_schema.has_field(nest_under):
         raise ValueError(f'The `nest_under` column {nest_under} does not exist.')
 
-      # Make sure the input path has the same cardinality as the nest_under path.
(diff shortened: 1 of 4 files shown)


nsthorat (approved):
This is awesome, only one high-level API comment.

I was thinking about this API from your description:

```
dataset.map(geo_locate, input_path='people.*.address', nest_under='people.*', output_column='geo')
```

And it made me think we should just merge nest_under and output_column, so you can just do this:

```
dataset.map(geo_locate, input_path='people.*.address', output_column='people.*.geo')
```

This makes my head spin a little bit less ^^

dsmilkov:
sweet, i like that, i can do it in a follow up since it will touch a bunch of files, and it should be purely mechanical

PR #1000, opened 2023-12-23
Make `ds.cluster()` have resumable title generation

dsmilkov:
Generating titles for the 9k clusters in openorca takes ~6-7hrs given the current rate limits and having resumability is important.

Also this brings us much closer to eliminating calling HDBScan as a signal and instead chaining multiple `ds.transform()` calls in a follow up PR

### DIFF
--- a/lilac/batch_utils.py
+++ b/lilac/batch_utils.py
@@ -86,7 +86,7 @@ def group_by_sorted_key_iter(
     key_fn: A function that takes an item and returns a key.
 
   Yields:
-    Each group of items sharing the same key.
+    A list of items sharing the same key, for each group.
   """
   last_key: object = None
   last_group: list[Item] = []
--- a/lilac/batch_utils.py
+++ b/lilac/batch_utils.py
@@ -76,6 +76,33 @@ TFlatBatchedInput = TypeVar('TFlatBatchedInput')
 TFlatBatchedOutput = TypeVar('TFlatBatchedOutput')
 
 
+def group_by_sorted_key_iter(
+  input: Iterator[Item], key_fn: Callable[[Item], object]
+) -> Iterator[list[Item]]:
+  """Takes a key-sorted iterator and yields each group of items sharing the same key.
+
+  Args:
+    input: An iterator of items, ordered by the key.
+    key_fn: A function that takes an item and returns a key.
+
+  Yields:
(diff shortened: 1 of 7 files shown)


lilac/batch_utils.py
```diff
@@ -76,6 +76,33 @@ def unflatten(
...
+  input: Iterator[Item], key_fn: Callable[[Item], object]
+) -> Iterator[list[Item]]:
+  """Takes a key-sorted iterator and yields each group of items sharing the same key.
+
+  Args:
+    input: An iterator of items, ordered by the key.
+    key_fn: A function that takes an item and returns a key.
+
+  Yields:
+    Each group of items sharing the same key.
```
nsthorat, line 89: nit: s/Each/A list of a

lilac/data/dataset.py
```diff
@@ -483,43 +483,30 @@ def cluster(
...
-        clusters[cluster_id] = docs
-
-      # Maps a cluster id to a topic.
-      cluster_topics: dict[str, str] = {}
-      # Compute the topic for each cluster.
-      for cluster_id, docs in tqdm(
-        list(clusters.items()), dynamic_ncols=True, desc='Generating topics'
-      ):
+    def _transform(items: Iterator[Item]) -> Iterator[Item]:
+      groups = group_by_sorted_key_iter(items, lambda x: x[signal_key][0]['cluster_id'])
```
nsthorat, line 487: can you make 'cluster_id' and 'membership_prob' constants?

lilac/data/dataset.py
```diff
@@ -483,43 +483,30 @@ def cluster(
     self.compute_signal(signal, path, overwrite=overwrite)
```
nsthorat, line 483: comment on line 449: Can we make this implementation in a separate file? It will be nice for us to see small pipelines as utilities as templates for further stuff

nsthorat (approved):

lilac/data/dataset.py
dsmilkov: done, moved to clustering.py, had to make dataset.cluster an abstract method due to circular dep, so now duckdb impl uses that outside util.

lilac/data/dataset.py
dsmilkov: done

PR #1002, opened 2023-12-26
`dataset.cluster()` now uses `transform()` which uses `map()`

dsmilkov:
Make `dataset.cluster()` use `transform()` which uses `map()` for the entire pipeline.

This completes the refactor needed so that signals can call `map()` instead.

- `dataset.cluster()` now writes the output to a sibling column (next to the original path, not under it)
- `dataset.map()` can now take optional schema. This is needed so that we can tag `cluster_id `as categorical. And it will unlock signals to call `map()`.
- `dataset.map()` can now take `embedding:str`. When provided, the map function will be called with the embedding spans for that item. This is needed for any computation that needs the embeddings (e.g. clustering). And it will unlock any `VectorSignal`s to call `map()` inside `compute_vector()`.
- `combine_columns` is now gone from `map()`. It will always combine, which is simpler since we will always write to a sibling
- Add utils for determining the sibling path and the common ancestor since we sometimes need our `map()` to see both the original info and the sibling info at once.


lilac/data/dataset_utils.py
```diff
@@ -346,3 +361,39 @@ def _cardinality_prefix(path: PathTuple) -> PathTuple:
...
+  """Get the output path for a sibling column."""
+  # Find the last non-wildcard in the path.
+  index = 0
+  for i, path_part in enumerate(path):
+    if path_part != PATH_WILDCARD:
+      index = i
+  return (*path[:index], f'{path[index]}_{suffix}', *path[index + 1 :])
+
+
+def get_ancestor_path(path1: PathTuple, path2: PathTuple) -> tuple[Optional[PathTuple], str, str]:
```
nsthorat, line 376: get_common_ancestor? wasnt clear from this title what it was doing

lilac/data/clustering.py
```diff
@@ -100,25 +97,49 @@ def cluster(
...
+    # Providing schema to avoid inferring and to flag the cluster_id as categorical so the histogram
+    # is sorted by size in the UI.
+    schema=field(fields={CLUSTER_ID: field('int32', categorical=True), MEMBERSHIP_PROB: 'float32'}),
+    overwrite=overwrite,
+  )
 
-  # Now that we have the clusters, compute the topic for each cluster with a map.
-  def _transform(items: Iterator[Item]) -> Iterator[Item]:
-    groups = group_by_sorted_key_iter(items, lambda x: x[signal_key][0][_CLUSTER_ID])
+  def compute_topics(
```
nsthorat, line 128: nit _compute_topics

lilac/data/dataset.py
```diff
@@ -801,13 +801,17 @@ def transform(
       sort_by: The path to sort by. If specified, the map will be called with rows sorted by this
         path. This is useful for map functions that need to maintain state across rows.
       sort_order: The sort order. Defaults to ascending.
+      embedding: The embedding to use for the transform function. If specified, the transform
```
nsthorat, line 804: is it called only with the embedding? if so use "only"

lilac/data/dataset_map_test.py
```diff
@@ -144,7 +144,7 @@ def test_map_signal(
   dataset.compute_signal(signal, 'text')
 
   def _map_fn(item: Item) -> Item:
-    return {'result': f'{item["text.test_signal.firstchar"]}_{len(item["text"])}'}
+    return {'result': f'{item["text"]["test_signal"]["firstchar"]}_{len(item["text"][VALUE_KEY])}'}
```
nsthorat, line 147: i think you know this but we should get rid of VALUE_KEY if combine_cols=True is the default now before the next release to not break anyone

lilac/data/dataset_map_test.py
```diff
@@ -858,20 +855,19 @@ def _map_fn(item: Item) -> Item:
   ]
 
 
-def test_map_combine_columns_with_input_path(make_test_data: TestDataMaker) -> None:
+def test_map_over_enriched_item(make_test_data: TestDataMaker) -> None:
```
nsthorat, line 858: this looks the same as the other test now

lilac/data/dataset_duckdb.py
```diff
@@ -623,23 +625,29 @@ def _get_vector_db_index(self, embedding: str, path: PathTuple) -> VectorDBIndex
...
     self,
     select_path: Optional[PathTuple] = None,
-    combine_columns: bool = False,
     resolve_span: bool = False,
     shard_cache_filepath: Optional[str] = None,
     overwrite: bool = False,
     query_options: Optional[DuckDBQueryParams] = None,
     shard_id: Optional[int] = None,
     shard_count: Optional[int] = None,
+    empty: bool = False,
```
nsthorat, line 634: what does empty mean? can we rename this to something more meaningful

lilac/data/clustering.py
```diff
@@ -132,12 +153,20 @@ def _transform(items: Iterator[Item]) -> Iterator[Item]:
...
         yield topic
 
-  output_path = output_path or (*path, 'topic')
+  # Now that we have the clusters, compute the topic for each cluster with another transform.
+  # The transform needs to be see both the original text and the cluster enrichment, so we need
+  # to map over the ancestor path.
+  ancestor_path, text_column, cluster_column = get_ancestor_path(path, cluster_output_path)
+
+  # Output the topic as a child of the cluster enrichment.
+  topic_output_path = (*cluster_output_path, 'topic')
```
nsthorat, line 162: topic => const

lilac/data/dataset.py
```diff
@@ -756,6 +754,10 @@ def map(
       execution_type: The local execution type of the map. Either "threads" or "processes". Threads
         are better for network bound tasks like making requests to an external server, while
         processes are better for CPU bound tasks, like running a local LLM.
+      embedding: If specified, the map function will be called with the embedding spans for that
```
nsthorat, line 757: it's worth pointing to documentation for what an embedding span is here (we call them VectorSpan so it will be hard for a user to find what this is)

nsthorat (approved):
nice work

lilac/data/clustering.py
dsmilkov: changed, in general i only use underscore if the fn is top level , since python makes it public api. internal functions are private by default.

lilac/data/dataset_map_test.py
dsmilkov: Yeah, VALUE_KEY will just disappear from any code once we make signals write to a sibling , instead of under a petal. This test calls a signal so i temporarily needs that VALUE_KEY until we fix the signal.

lilac/data/dataset.py
dsmilkov: done

lilac/data/dataset_duckdb.py
dsmilkov: i removed it -- it was an optimization for selecting only the arbitrarily nested list of nulls w/o the text, just like we do with signal.vector_compute() since we don't need the text.

lilac/data/dataset_map_test.py
dsmilkov: removed

PR #1004, opened 2023-12-26
Add signal.map customization

brilee:
Some design ideas:

Constraint map_fn to just take a plain Callable, instead of Union[Callable, Signal], which means we would remove the Embedding/VectorSignal/Signal branches inside dataset._dispatch_workers. I'm also hoping to remove the other ad-hoc  batch/map/parallel implementation that lives inside embedding.compute_split_embeddings, and get everything on top of the same map API.

Difficulties... Signal.compute becomes polymorphic - it could have a signature of Item -> Item, List[Item] -> List[Item], or Iterator[Item] -> Iterator[Item] depending on whether Signal.map_batch_size is -1, int, or None.

I also have no idea why VectorSignal.vector_compute looks so complicated...

The above sequence would be a breaking change for users, but I suspect it will remove a lot of complexity from their code so it'd be welcome.

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -1027,11 +1027,13 @@ class DatasetDuckDB(Dataset):
     _consume_iterator(
       progress_bar(
         self._dispatch_workers(
-          joblib.Parallel(n_jobs=1, backend='threading', return_as='generator'),
+          joblib.Parallel(
+            n_jobs=signal.map_parallelism, backend=signal.map_strategy, return_as='generator'
+          ),
           signal,
           output_path,
           jsonl_cache_filepath,
-          batch_size=None,
+          batch_size=signal.map_batch_size,
           select_path=input_path,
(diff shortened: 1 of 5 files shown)


dsmilkov (approved):
+1 to the refactor since now is the time to do the breaking (including signals writing output to a sibling) before we release 0.2.x

nsthorat:
+1 to refactor, though it's nice for mao to take a signal in practice. Signals provide a bit more metadata that gets written in the signal dictionary on the manifest for the UI (it really should be a signal, not an untyped dictionary, that was a mistake from a while ago).

nsthorat (approved):

brilee:
agreed - compute_signal shouldn't dispatch directly to map(), but will go through _dispatch_workers, so that it can continue to keep track of signal-related metadata. I only intend to remove the special casing of the signal.compute inside _dispatch_workers.

nsthorat:
that makes a lot of sense

PR #1005, opened 2023-12-27
Make OpenAI calls threaded with exponential backoff

dsmilkov:
We can now summarize ~15 clusters / secs. Slimorca 10k sample has 250 clusters, and is done in ~16 secs.

### DIFF
--- a/lilac/data/clustering.py
+++ b/lilac/data/clustering.py
@@ -3,9 +3,11 @@ import functools
 from typing import Any, Iterator, Optional
 
 import instructor
+from joblib import Parallel, delayed
 from pydantic import (
   BaseModel,
 )
+from tenacity import retry, stop_after_attempt, wait_random_exponential
 
 from ..batch_utils import group_by_sorted_key_iter
 from ..schema import (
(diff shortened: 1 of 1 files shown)


lilac/data/clustering.py
```diff
@@ -127,32 +131,41 @@ def _compute_clusters(span_vectors: Iterator[list[SpanVector]]) -> Iterator[Item
...
-        docs.append((text, membership_prob))
-
-      # Sort by membership score.
-      sorted_docs = sorted(docs, key=lambda x: x[1], reverse=True)
-      topic = topic_fn(sorted_docs) if sorted_docs else None
-
-      # Yield a topic for each item in the group since the combined output needs to be the same
-      # length as the combined input.
-      for item in group:
+    parallel = Parallel(n_jobs=_NUM_THREADS, backend='threading', return_as='generator')
```
nsthorat, line 161: presumably it doesn't make sense to reuse map() threads?

nsthorat (approved):

lilac/data/clustering.py
brilee: don't think it does... I think this situation is probably weird enough that we can special case it for now regardless of any other signal API fixes I'm working on.

dsmilkov: Here we are already inside a map() through a transform, so this is now internal detail.

PR #1006, opened 2023-12-28
Implement signals on top of the map infrastructure

brilee:
I have not figured out what to do with VectorSignals - it seems like there's some duplication of functionality with the `embedding` argument to map(). So those are still special-cased.

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -974,7 +974,15 @@ class DatasetDuckDB(Dataset):
     task_id: Optional[TaskId] = None,
   ) -> None:
     if isinstance(signal, TextEmbeddingSignal):
-      return self.compute_embedding(signal.name, path, overwrite=overwrite, task_id=task_id)
+      return self.compute_embedding(
+        signal.name,
+        path,
+        filters=filters,
+        limit=limit,
+        include_deleted=include_deleted,
+        overwrite=overwrite,
+        task_id=task_id,
+      )
 
     input_path = normalize_path(path)
 
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -744,7 +744,7 @@ class DatasetDuckDB(Dataset):
   def _dispatch_workers(
     self,
     pool_map: joblib.Parallel,
-    transform_fn: Union[Signal, Callable[[Iterable[Item]], Iterator[Optional[Item]]]],
+    transform_fn: Union[VectorSignal, Callable[[Iterable[Item]], Iterator[Optional[Item]]]],
     output_path: PathTuple,
     jsonl_cache_filepath: str,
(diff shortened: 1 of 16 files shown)


lilac/data/dataset_duckdb.py
```diff
@@ -1030,7 +1032,7 @@ def compute_signal(
           joblib.Parallel(
             n_jobs=signal.map_parallelism, prefer=signal.map_strategy, return_as='generator'
           ),
-          signal,
+          signal if isinstance(signal, VectorSignal) else signal.compute,
```
dsmilkov, line 1035: if you change `vectorsignal.vector_compute` to take `iterable[list[SpanVector]`, then you can pass

`pool_map=signal.vector_compute` and `embedding=True` if `isinstance(signal, VectorSignal)` and delete the special code path in dispatch_workers

lilac/embeddings/embedding.py
```diff
@@ -65,66 +63,25 @@ def _embed_fn(data: Iterable[RichData]) -> Iterable[list[SpanVector]]:
...
 
-def compute_split_embeddings(
-  docs: Iterable[str],
-  batch_size: int,
+def identity_chunker(doc: str) -> list[TextChunk]:
+  """Split a document into a single chunk."""
+  return [(doc, (0, len(doc)))]
+
+
+def chunked_compute_embedding(
```
dsmilkov, line 71: this is great, just for my own understanding: I guess there is a slightly perf penalty in that if the batch size is 96 docs, and those 96 docs have 100 chunks total, then we will call embed_fn once with 96 and again with 4 chunks, the 2nd call is suboptimal, but that's ok since all the complexity of unpacking the 2nd batch went away.

dsmilkov (approved):
nice cleanup!

nsthorat (approved):

lilac/embeddings/embedding.py
brilee: that's right; the map api has no idea that the map_fn is internally splitting up the work even further, so there's not much we can do about that here. This impl will be 50-100% as efficient as the original, depending on how much the chunker splits up stuff.

lilac/signal.py
```diff
@@ -94,8 +93,25 @@ def fields(self) -> Optional[Field]:
     """
     return None
 
-  def compute(self, data: Iterable[RichData]) -> Iterator[Optional[Item]]:
-    """Compute the signal for an iterable of documents or images.
+  def compute(self, data: Any) -> Any:
```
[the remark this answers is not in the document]
brilee: Tried; mypy throws all sorts of Liskov Substitution Principle violations that I don't really understand... :/

lilac/data/dataset_duckdb.py
brilee: let me check this in and then fix in a followup CL

PR #1007, opened 2023-12-28
Allow clustering of a nested path

dsmilkov:
Clustering of `something.*.*` is not yet supported. The path must end in a field name for now.

### DIFF
--- a/lilac/data/cluster_test.py
+++ b/lilac/data/cluster_test.py
@@ -1,4 +1,5 @@
 """Unit tests for dataset.cluster()."""
+import re
 from typing import Iterable
 
 import pytest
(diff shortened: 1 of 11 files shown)


lilac/data/cluster_test.py
```diff
@@ -59,3 +60,69 @@ def topic_fn(docs: list[tuple[str, float]]) -> str:
...
+          'text': 'Can you rewrite this in a simpler way',
+          'text_cluster': {'cluster_id': 1, 'membership_prob': 1.0, 'topic': 'simplification'},
+        },
+      ]
+    },
+    {
+      'texts': [
+        {
+          'text': 'Can you provide a short summary of the following text',
+          'text_cluster': {'cluster_id': 0, 'membership_prob': 1.0, 'topic': 'summarization'},
```
nsthorat, line 101: for my understanding where does "_cluster" come from? maybe this should be `text.cluster` if auto-gen

nsthorat (approved):
sweet!

lilac/data/cluster_test.py
dsmilkov: let's avoid having periods because that makes it unclear if `cluster` is a child of `text`, or a column name with a dot in it. If we do a "dot", we will now forever have to escape it with double quotes in order to select that field in text form
`select_rows(['texts.*."text.cluster"'])` vs simply `select_rows(['texts.*.text_cluster'])`. Obviously a tuple will work, but the correct tuple will be tricky to write when using public api in notebooks.

PR #1008, opened 2023-12-29
remove VectorCompute path in dispatch_workers

brilee:
Breaking change: VectorSignal.vector_compute takes a set of spanvectors now, instead of a set of pathkeys and a vectordb


dsmilkov (approved):
nice cleanup

PR #1009, opened 2024-01-02
Add Jina (Small) on Garden signal

dsmilkov:


lilac/embeddings/jina_garden.py
```diff
@@ -0,0 +1,49 @@
...
+
+from ..schema import Item, lilac_embedding
+from ..signal import TextEmbeddingSignal
+from ..tasks import TaskExecutionType
+
+JINA_CONTEXT_SIZE = 8192
+
+
+class JinaV2SmallGarden(TextEmbeddingSignal):
+  """Jina V2 Embeddings with 8K context, hosted on Garden.
```
nsthorat, line 19: maybe we should use "Lilac Garden" when we say Garden

lilac/embeddings/jina_garden.py
```diff
@@ -0,0 +1,49 @@
...
+from lilac.batch_utils import compress_docs
+
+from ..schema import Item, lilac_embedding
+from ..signal import TextEmbeddingSignal
+from ..tasks import TaskExecutionType
+
+JINA_CONTEXT_SIZE = 8192
+
+
+class JinaV2SmallGarden(TextEmbeddingSignal):
```
nsthorat, line 18: in the background we should think about whether garden should just become a bit to the embedding (it's not too important if you compute this on garden or locally, as long as the outputs are the same)

nsthorat (approved):

lilac/embeddings/jina_garden.py
dsmilkov: done

lilac/embeddings/jina_garden.py
dsmilkov: Yeah, i like that. we can re-work this later easily... we can also add that bit to the dataset.cluster

PR #1010, opened 2024-01-03
Add `dataset.cluster(remote=True)` bit

dsmilkov:
Also remove the now oudated HDBScan and DBScan signals


lilac/data/dataset.py
```diff
@@ -451,6 +451,7 @@ def cluster(
     min_cluster_size: int = 5,
     topic_fn: Optional[TopicFn] = None,
     overwrite: bool = False,
+    remote: bool = False,
```
nsthorat, line 454: I really like this pattern. I think we should long-term do the same for embeddings & signals, instead of creating copies!

nsthorat (approved):
amazingn work dude

PR #1026, opened 2024-01-05
Allows non folder exports

hynky1999:
I was unable to export the dataset from UI, when I only specified the name of the file.
The problem. is that python can't create empty directories, but if we don't specify one the code still attempts to create a directory. This will cause an error.

This PR fixes it


dsmilkov (approved):
Thank you!

nsthorat (approved):

PR #1031, opened 2024-01-05
Fix some `map(overwrite=True)` bugs

dsmilkov:
Fix progress bar when overwrite=True and allow `map(input=x, output=x, overwrite=True)`

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -609,8 +609,10 @@ class DatasetDuckDB(Dataset):
       self._vector_indices[index_key] = vector_index
       return vector_index
 
-  def _get_cache_len(self, cache_filepath: str) -> int:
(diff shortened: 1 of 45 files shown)


lilac/data/dataset_duckdb.py
```diff
@@ -2664,17 +2665,6 @@ def map(
           field = manifest.data_schema.get_field(output_path)
           if field.map is None:
             raise ValueError(f'{output_path} is not a map column so it cannot be overwritten.')
-          # Delete the parquet file and map manifest.
```
dsmilkov, line 2667: we don't need this code since the parquet file and the map manifest are deleted/overwritten when we reshard. This makes map(input=x, output=x, overwrite=True) work

nsthorat (approved):
nice, __ is much better

PR #1044, opened 2024-01-08
Fix bug with ItemMedia not rendering media fields that are deeply nested siblings.

nsthorat:
https://lilacai-nikhil-staging.hf.space/datasets#local/Capybara

image

Click around other datasets.

Also:
- Fix bug with limit of select_groups from the pivot viewer.
- Removes some dead datasts in the demo config.

### DIFF
--- a/web/blueprint/src/lib/components/datasetView/DatasetPivotResult.svelte
+++ b/web/blueprint/src/lib/components/datasetView/DatasetPivotResult.svelte
@@ -12,7 +12,7 @@
   import Carousel from 'svelte-carousel';
 
   import {datasetLink} from '$lib/utils';
-  import {ROWID, serializePath, type BinaryFilter, type Path, type UnaryFilter} from '$lilac';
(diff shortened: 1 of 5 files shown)


web/blueprint/src/lib/components/datasetView/DatasetPivotResult.svelte
```diff
@@ -53,7 +53,9 @@
   $: countQuery = shouldLoad
     ? querySelectGroups($store.namespace, $store.datasetName, {
         leaf_path: path,
-        filters
+        filters,
```
dsmilkov, line 56: looks like parentPath above is an unused exported prop

dsmilkov (approved):
nice!

web/blueprint/src/lib/components/datasetView/DatasetPivotResult.svelte
nsthorat: done

PR #1049, opened 2024-01-09
Add remote computation for the jina embedding

dsmilkov:


lilac/signal.py
```diff
@@ -77,6 +77,8 @@ class Signal(BaseModel):
   map_parallelism: ClassVar[int] = 1
```
nsthorat, line 77: local_batch_size / local_parallelism / local_strategy?

nsthorat (approved):

lilac/signal.py
dsmilkov: done

PR #1052, opened 2024-01-10
Add indexing on database startup, flag-guarded

brilee:
Flag is hardcoded now, haven't figured out the .env thing yet

### DIFF
--- a/lilac/batch_utils.py
+++ b/lilac/batch_utils.py
@@ -139,6 +139,6 @@ def compress_docs(docs: list[str]) -> bytes:
   """Compress a list of documents using gzip."""
   buffer = BytesIO()
   # Gzip compress the data and write it to the buffer
-  with gzip.GzipFile(fileobj=buffer, mode='wb') as f:
+  with gzip.GzipFile(fileobj=buffer, mode='wb', compresslevel=6) as f:
     f.write(json.dumps(docs).encode('utf-8'))
   return buffer.getvalue()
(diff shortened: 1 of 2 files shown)


lilac/batch_utils.py
```diff
@@ -139,6 +139,6 @@ def compress_docs(docs: list[str]) -> bytes:
   """Compress a list of documents using gzip."""
   buffer = BytesIO()
   # Gzip compress the data and write it to the buffer
-  with gzip.GzipFile(fileobj=buffer, mode='wb') as f:
+  with gzip.GzipFile(fileobj=buffer, mode='wb', compresslevel=6) as f:
```
dsmilkov, line 142: just curious: is 6 just a good speed/size tradeoff?

lilac/data/dataset_duckdb.py
```diff
@@ -184,6 +184,9 @@
   'less_equal': '<=',
 }
 
+USE_TABLE_INDEX = True
```
dsmilkov, line 187: we should make this an env flag

dsmilkov (approved):

nsthorat (approved):
looks good -- lets do env flag before submitting

lilac/batch_utils.py
brilee: yep, default is 9 which is pretty slow; 6 is something like 2-10x faster with only a 10-25% larger file

lilac/data/dataset_duckdb.py
brilee: done

Issue #1047, opened 2024-01-09
OpenAI Authentication error when computing embeddings via python API


asanaa8:
Hello, I'm getting an authentication error code 401 when computing embeddings  via python API. The error mentionned that I have an Incorrect API Key. Actually, I used the same API key to compute openai embeddings via UI and it works. 

The used env variables are:
``
export OPENAI_API_KEY=...
export OPENAI_API_TYPE=azure
export OPENAI_API_VERSION=2022-12-01
export AZURE_OPENAI_ENDPOINT=https://{...}.openai.azure.com/
``
![1](https://github.com/lilacai/lilac/assets/93333719/c40537e7-b932-4c1e-b62b-2feb7ddcab76)

asanaa8:
If I remove these env variables, lilac will ask for them so it doesn't work without. OPENAI_API_KEY, OPENAI_API_VERSION, OPENAI_API_TYPE, OPENAI_API_BASE, OPENAI_API_ENGINE_EMBEDDING are mondatory to let lilac call azure openai api. It works fine for me via UI but when using Python API, NO :'(

I'm on lilac 0.2.2

nsthorat:
@asanaa8 this should be fixed in 0.2.3 -- please reopen if you still have this issue!

PR #1053, opened 2024-01-10
fix: migrate embeddings by azure openai to openai > 1.0.0

dechantoine:
Hey ! A short PR to fix this issue : https://github.com/lilacai/lilac/issues/1047
I chose to raise an error when both OPENAI_API_KEY and AZURE_OPENAI_KEY environment variables are set because I think it's better to not default to one without warning the user.

I tested OpenAI API & AzureAPI on both UI and Python and everything works like charm :)

### DIFF
--- a/lilac/embeddings/openai.py
+++ b/lilac/embeddings/openai.py
@@ -32,17 +32,23 @@ class OpenAIEmbedding(TextEmbeddingSignal):
 
   name: ClassVar[str] = 'openai'
   display_name: ClassVar[str] = 'OpenAI Embeddings'
-  local_batch_size: ClassVar[int] = API_OPENAI_BATCH_SIZE
-  local_parallelism: ClassVar[int] = API_NUM_PARALLEL_REQUESTS
+  local_batch_size: ClassVar[int] = None
+  local_parallelism: ClassVar[int] = None
   local_strategy: ClassVar[TaskExecutionType] = 'threads'
 
   @override
   def setup(self) -> None:
     api_key = env('OPENAI_API_KEY')
-    api_type = env('OPENAI_API_TYPE')
-    api_version = env('OPENAI_API_VERSION')
-    if not api_key:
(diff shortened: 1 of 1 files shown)


brilee:
https://github.com/lilacai/lilac/pull/1052 has the fix for the flaky test -  ```FAILED lilac/data/cluster_test.py::test_simple_clusters[DatasetDuckDB] - duckdb.duckdb.IOException: IO Error: No files found that match the pattern "/tmp/pytest-of-runner/pytest-...```, submitting that PR in just a moment.

lilac/embeddings/openai.py
```diff
@@ -32,17 +32,25 @@ class OpenAIEmbedding(TextEmbeddingSignal):
 
   name: ClassVar[str] = 'openai'
   display_name: ClassVar[str] = 'OpenAI Embeddings'
-  local_batch_size: ClassVar[int] = API_OPENAI_BATCH_SIZE
-  local_parallelism: ClassVar[int] = API_NUM_PARALLEL_REQUESTS
+  local_batch_size: ClassVar[int] = None
```
brilee, line 35: I think this will throw mypy typechecker errors due to the non-optional type annotation. Can we leave placeholder integer values here?

brilee (approved):

nsthorat (approved):
Awesome work, +1 to Brian's comment and then we'll merge it.

Appreciate it!

lilac/embeddings/openai.py
dechantoine: Thank you for the correction ! Should we put a dummy value or leave API_OPENAI_BATCH_SIZE & API_NUM_PARALLEL_REQUESTS ?

nsthorat: I think the constants are great!

dechantoine:
There you go :)

nsthorat:
Thanks @dechantoine

PR #1055, opened 2024-01-11
Add a notebook for working with concepts from python.

nsthorat:


dsmilkov (approved):

notebooks/UsingConcepts.ipynb
```diff
@@ -0,0 +1,268 @@
+{
```
brilee, line 1: <p>could you print the shape of the coefficients and make clear that the coefficients are linked to the specific embedding used? Can also demonstrate that you can use a custom embedding/model fn.</p><p> </p><p>Also, some text about how the magic is all in the specific choice of positive/negative examples, and that lilac's UI is a powerful way to craft these example sets.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lilacai/lilac/pull/1055/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

PR #1058, opened 2024-01-12
Improve the cluster titling and fix a few client-side bugs

dsmilkov:
- use sort_by cluster_id so we can avoid any thread locks and get better resumability
- Improve the prompts
- Fix a few client-side and server-side bugs related to sorting
- Title-ing is now 2x faster -- takes 10mins to title HN 1M comments, and 13min to title OpenHermes-2.5

### DIFF
--- a/lilac/data/clustering.py
+++ b/lilac/data/clustering.py
@@ -360,6 +360,7 @@ def cluster(
         task_info=task_info,
       )
       for item, title in zip(items2, titles):
+        # Drop the temporary newline-concatenated text column.
         del item[TEXT_COLUMN]
         yield {**item, CATEGORY_TITLE: title}
 
@@ -398,7 +399,6 @@ def _cluster(
   """Cluster docs with HDBSCAN."""
   if remote:
     remote_fn = modal.Function.lookup('cluster', 'Cluster.cluster').remote
-    docs = list(docs)
     gzipped_docs = compress_docs(list(docs))
     response = remote_fn({'gzipped_docs': gzipped_docs})
     yield from response['clusters']
--- a/lilac/data/clustering.py
+++ b/lilac/data/clustering.py
@@ -315,7 +315,7 @@ def cluster_impl(
 
     def compute_clusters(items: Iterator[Item]) -> Iterator[Item]:
       items, items2 = itertools.tee(items)
-      docs: Iterator[Optional[str]] = (item[TEXT_COLUMN] for item in items)
+      docs: Iterator[Optional[str]] = (item.get(TEXT_COLUMN) for item in items)
       cluster_items = sparse_to_dense_compute(
         docs, lambda x: _hdbscan_cluster(x, min_cluster_size, remote)
       )
--- a/lilac/data/clustering.py
+++ b/lilac/data/clustering.py
@@ -1,8 +1,8 @@
 """Clustering utilities."""
 import functools
 import gc
+import itertools
 import random
-import threading
 from typing import Any, Callable, Iterator, Optional, Union, cast
 
 import instructor
@@ -14,7 +14,7 @@ from pydantic import (
 )
 from tenacity import retry, stop_after_attempt, wait_random_exponential
 
-from ..batch_utils import compress_docs, flatten_path_iter
+from ..batch_utils import compress_docs, flatten_path_iter, group_by_sorted_key_iter
 from ..embeddings.jina import JinaV2Small
 from ..schema import (
   EMBEDDING_KEY,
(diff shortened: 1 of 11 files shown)


lilac/data/clustering.py
```diff
@@ -150,22 +200,28 @@ def cluster(
   topic_fn: TopicFn = summarize_request,
   overwrite: bool = False,
   remote: bool = False,
-  category: bool = False,
   task_id: Optional[TaskId] = None,
+  recompute_titles: bool = False,
```
dsmilkov, line 204: note: this param is not public API, clustering.cluster() is an internal util that dataset.cluster calls

lilac/data/clustering.py
```diff
@@ -94,15 +94,16 @@ def summarize_request(ranked_docs: list[tuple[str, float]]) -> str:
...
-    top_p=0.1,
     max_tokens=50,
     messages=[
       {
         'role': 'system',
         'content': (
-          'Ignore the group of requests below, and create a 4-5 word title to '
-          'summarize the whole group. Some examples: "Classifying book review sentiment", '
-          '"Questions about South East Asia", "Translating English to Polish", etc.'
+          'You are a world-class title generator. Ignore the group of related requests below, and '
```
nsthorat, line 102: lol

lilac/data/clustering.py
nsthorat: maybe we should rename the method then (cluster_impl or something)

web/blueprint/src/lib/queries/datasetQueries.ts
```diff
@@ -174,10 +174,14 @@ function getRowMetadataBatcher(
       fetcher: async (request: BatchMetadataRequest[]) => {
         const rowIds = request.map(r => r.rowId);
         const selectRowsResponse = await DatasetsService.selectRows(namespace, datasetName, {
-          filters: [{path: [ROWID], op: 'in', value: rowIds}],
+          // Add a IN filter for the rowIds, in addition to any other existing filters.
```
nsthorat: instead of explaining the line of code, maybe something like "we add the filters to improve the performance of the query by restricting the set of rows"

nsthorat (approved):

lilac/data/clustering.py
brilee: is it any surprise that aproduct coming out of silicon valley works best when you speak to it like an overconfident and clueless CEO?

lilac/data/clustering.py
```diff
@@ -185,205 +241,151 @@ def cluster(
...
       return '\n'.join(texts)
 
-    map_fn = input if callable(input) else flatten_input
-    dataset.map(map_fn, output_path=path, overwrite=overwrite)
+    def extract_text(item: Item) -> Item:
+      cluster_info = item
+      for path_part in cluster_output_path:
+        cluster_info = cluster_info.get(path_part, {})
+
+      text = input(item) if callable(input) else _flatten_input(item, cast(PathTuple, path))
```
brilee, line 265: it's a little confusing that the arg name "input" collides with a python builtin

lilac/data/clustering.py
```diff
@@ -185,205 +241,151 @@ def cluster(
...
-    overwrite=overwrite,
-    remote=remote,
-    category=True,
-    task_id=task_id,
-  )
+  category_id_exists = schema.has_field((*cluster_output_path, CATEGORY_ID))
+  if not category_id_exists or overwrite or recompute_titles:
+    if task_info:
+      task_info.message = 'Computing super clusters'
+      task_info.total_progress = 0
```
brilee, line 323: Does the task_id need to get passed to the transform calls in order to provide progress bars?

brilee (approved):
I love the rewriting final column strategy!

lilac/data/clustering.py
dsmilkov: done

lilac/data/clustering.py
dsmilkov: loooool

web/blueprint/src/lib/queries/datasetQueries.ts
dsmilkov: this was reverted -- didn't need other filters, since this is just metadata exact rowid lookup

lilac/data/clustering.py
dsmilkov: yes!

lilac/data/clustering.py
dsmilkov: renamed to input_fn_or_path, better to be explicit and this is internal util

PR #1059, opened 2024-01-12
Add an upload dataset script. Some other cleanups.

nsthorat:
Python:
- Add limit, offset, filter to labels & deleting. This allows us to filter by a cluster group, sort by membership prob, limit=N, and add a label to it
- Export useful methods to lilac public API.

UI:
- Remove filters, searches, group by when clicking the cluster button in the UI

Example dataset: https://huggingface.co/datasets/lilacai/hermes-cluster-sample

image

### DIFF
--- a/lilac/__init__.py
+++ b/lilac/__init__.py
@@ -11,7 +11,7 @@ from .config import (
 )
 from .data import *  # noqa: F403
 from .data.dataset_duckdb import DatasetDuckDB
-from .db_manager import get_dataset, set_default_dataset_cls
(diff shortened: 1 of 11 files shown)


dsmilkov (approved):

brilee (approved):
re: sort/limiting: does it make more sense to filter by membership prob > 0.5?

nsthorat:
It could make sense, but I'm actually just choosing the top-k sorted by membership prob desc for the prototype ipynb (so the filter doesnt matter).

But he could decide to do that if not targeting a specific size output dataset.

PR #1063, opened 2024-01-12
Cluster spec deployer

brilee:
I haven't fully tested this - specifically, I'm not sure if the "detect previously computed clusters" works.

Also add a random fix for Daniel for https://github.com/lilacai/lilac/pull/1062#pullrequestreview-1818978296

### DIFF
--- a/lilac/load.py
+++ b/lilac/load.py
@@ -2,9 +2,10 @@
 
 Usage:
 
-poetry run python -m lilac.load \
-  --project_dir=demo_data \
-  --config_path=lilac_hf_space.yml
+poetry run lilac load \
+  --config_path=lilac_hf_space.yml \
+  demo_data
+
 """
 
 import gc
@@ -61,6 +62,11 @@ def load(
   if not isinstance(config, Config):
     config_path = config or os.path.join(project_dir, PROJECT_CONFIG_FILENAME)
     config = read_config(config_path)
+
+  from devtools import pformat
+
+  print(pformat(config, highlight=False))
+  return
 
   task_manager = get_task_manager()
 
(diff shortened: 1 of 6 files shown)


lilac/load.py
```diff
@@ -192,6 +193,20 @@ def load(
...
       gc.collect()
 
+  log()
+  log('*** Compute clusters ***')
+  with DebugTimer('Computing clusters'):
+    for c in config.clusters:
+      dataset = DatasetDuckDB(c.dataset_namespace, c.dataset_name, project_dir=project_dir)
+      schema = dataset.manifest().data_schema
+      # Try to find the cluster if it is precomputed.
+      for node in schema.fields.values():
```
dsmilkov, line 203: use `schema.all_fields` so it can find it even if the result is nested a bit (can happen if we cluster (chapter, section, text), the result will be in (chapter, section, text_cluster)

dsmilkov (approved):

lilac/load.py
```diff
@@ -2,9 +2,10 @@
 
 Usage:
 
-poetry run python -m lilac.load \
-  --project_dir=demo_data \
-  --config_path=lilac_hf_space.yml
+poetry run lilac load \
```
nsthorat, line 5: we dont really use this ever, we could remove it

lilac/load.py
```diff
@@ -2,9 +2,10 @@
...
+  --config_path=lilac_hf_space.yml \
+  demo_data
+
 """
```
nsthorat, line 9: can you add a unit test for cluster in load_test.py? you could cheat and mock it, just make sure the config does the right thing

brilee: done

PR #1068, opened 2024-01-16
Improve the cluster/pivot UI

dsmilkov:
- Speedup rendering when we have lots of categories (full orca has ~700 categories)
- Move pivot calculation server-side (dataset.pivot())
- Enable fast, client-side search
- Fix a problem with deep-equal dep (shows up when you rm svelte-kit, and try to rebuild it)
- Fix a problem with incomplete output with openai
- We can now search 700 categories and 36k clusters in ms (Full open orca 4.3M rows)

https://github.com/lilacai/lilac/assets/2294279/7f1cc0b3-648c-488f-a68e-a9bdce71f38b


lilac/data/clustering.py
```diff
@@ -111,26 +112,32 @@ def summarize_request(ranked_docs: list[tuple[str, float]]) -> str:
...
-            '"Writing product descriptions", etc. Prefer using descriptive words. Do not use vague '
-            'words like "various", "assortment", "comments", "discussion", etc.'
-          ),
-        },
-        {'role': 'user', 'content': input},
-      ],
-    )
-    return title.title
+    max_tokens = 50
+    while True:
```
nsthorat, line 116: should this have a max tries so we dont while loop forever?

lilac/data/dataset.py
```diff
@@ -273,6 +273,21 @@ class SelectGroupsResult(BaseModel):
   bins: Optional[list[Bin]] = None
 
 
+class OuterGroup(BaseModel):
+  """The outer group of a pivot query."""
```
nsthorat, line 277: s/query/result

lilac/data/dataset.py
```diff
@@ -273,6 +273,21 @@ class SelectGroupsResult(BaseModel):
   bins: Optional[list[Bin]] = None
 
 
+class OuterGroup(BaseModel):
```
nsthorat, line 276: PivotResultOuterGroup

lilac/data/dataset.py
```diff
@@ -564,6 +579,18 @@ def select_groups(
...
+  @abc.abstractmethod
+  def pivot(
+    self,
+    outer_path: Path,
+    inner_path: Path,
+    filters: Optional[Sequence[FilterLike]] = None,
+    sort_by: Optional[GroupsSortBy] = GroupsSortBy.COUNT,
+    sort_order: Optional[SortOrder] = SortOrder.DESC,
+  ) -> PivotResult:
+    """Pivot table with counts across two fields (outer and inner)."""
```
nsthorat, line 591: can you make this docstring a bit more descriptive since it's public API?

Would be good to say something like:

```
Generate a pivot table over two fields.

For each value of outer_path in the dataset, the pivot table will contain the count of that value, as well as the distribution of all values of the inner_path, given the outer path.
```

Can you also give an example for users here, maybe something like splits='train'|'test' and source='wiki'|'c4' or something.

lilac/data/dataset_duckdb.py
```diff
@@ -1629,6 +1631,107 @@ def select_groups(
...
+    while outer_leaf.repeated_field:
+      outer_leaf = outer_leaf.repeated_field
+      outer_path = (*outer_path, PATH_WILDCARD)
+
+    if not inner_leaf.dtype:
+      raise ValueError(f'Inner path "{inner_path}" is not a leaf in the dataset')
+    if not outer_leaf.dtype:
+      raise ValueError(f'Outer path "{outer_path}" is not a leaf in the dataset')
+
+    if inner_leaf.dtype.type == 'map' or is_ordinal(inner_leaf.dtype):
```
nsthorat, line 1668: separate these to two throws since they're totally unrelated

lilac/router_dataset.py
```diff
@@ -257,6 +258,28 @@ def select_groups(
...
+  """The request for the pivot endpoint."""
+
+  inner_path: Path
+  outer_path: Path
+  filters: Sequence[Filter] = []
+
+
+@router.post('/{namespace}/{dataset_name}/pivot')
+def pivot(namespace: str, dataset_name: str, options: PivotOptions) -> PivotResult:
+  """Wrapper around dataset.pivot."""
```
nsthorat, line 271: how about REST endpoint for dataset.pivot

web/blueprint/src/lib/components/datasetView/DatasetPivotResult.svelte
```diff
@@ -1,42 +1,43 @@
...
 
   let isOnScreen = false;
   let root: HTMLDivElement;
   let observer = new IntersectionObserver(entries => {
     entries.forEach(entry => {
-      if (entry.isIntersecting) {
-        isOnScreen = true;
-        observer.disconnect();
-      }
+      isOnScreen = entry.isIntersecting;
```
nsthorat, line 40: do you need this anymore

web/blueprint/src/lib/components/datasetView/DatasetPivotViewer.svelte
```diff
@@ -55,30 +60,63 @@
...
-    const hasMatchingSearch = textHighlights.some(x => x.isBold);
-    return {
-      name,
-      count,
-      percentage: getPercentage(count),
-      textHighlights,
-      hasMatchingSearch
-    };
+
+  $: pivotQueryFn = createApiQuery(DatasetsService.pivot, DATASETS_TAG, {
```
nsthorat, line 72: this should be in datasetQueries.ts

lilac/data/dataset.py
```diff
@@ -564,6 +579,18 @@ def select_groups(
     """
     raise NotImplementedError
 
+  @abc.abstractmethod
+  def pivot(
```
nsthorat, line 583: probably worth a simple unit test here

nsthorat (approved):

lilac/data/clustering.py
```diff
@@ -111,26 +112,32 @@ def summarize_request(ranked_docs: list[tuple[str, float]]) -> str:
...
+                'Polish", "Writing product descriptions", etc. Prefer using descriptive words. Do '
+                'not use vague words like "various", "assortment", "comments", "discussion", etc.'
+              ),
+            },
+            {'role': 'user', 'content': input},
+          ],
+        )
+        return title.title
+      except IncompleteOutputException:
+        max_tokens = max_tokens * 2
```
brilee, line 139: 50 tokens is already a pretty generous limit for a topic name. I bet you could just add "Be concise." in a loop to the prompt 🤣

lilac/data/dataset_duckdb.py
```diff
@@ -1629,6 +1631,107 @@ def select_groups(
...
+      raise ValueError('both `outer_path` and `inner_path` must be provided')
+    sort_by = sort_by or GroupsSortBy.COUNT
+    sort_order = sort_order or SortOrder.DESC
+    inner_path = normalize_path(inner_path)
+    outer_path = normalize_path(outer_path)
+    manifest = self.manifest()
+    inner_leaf = manifest.data_schema.get_field(inner_path)
+    outer_leaf = manifest.data_schema.get_field(outer_path)
+
+    # Find the inner-most inner leaf in case this field is repeated.
```
brilee, line 1653: does it make sense to have a GROUP BY on a repeated field? I thought this should only apply to non-repeating leaf nodes

lilac/data/dataset_duckdb.py
```diff
@@ -1629,6 +1631,107 @@ def select_groups(
...
+      raise ValueError(f'Inner path "{inner_path}" is not a leaf in the dataset')
+    if not outer_leaf.dtype:
+      raise ValueError(f'Outer path "{outer_path}" is not a leaf in the dataset')
+
+    if inner_leaf.dtype.type == 'map' or is_ordinal(inner_leaf.dtype):
+      raise ValueError(f'Cannot compute pivot on an ordinal or map field "{inner_path}".')
+    if outer_leaf.dtype.type == 'map' or is_ordinal(outer_leaf.dtype):
+      raise ValueError(f'Cannot compute pivot on an ordinal or map field "{outer_path}".')
+
+    if self.stats(inner_path).approx_count_distinct >= dataset.TOO_MANY_DISTINCT:
```
brilee, line 1673: I think we should still pivot on the outer path if that one doesn't have too many distinct

lilac/data/dataset_duckdb.py
```diff
@@ -1629,6 +1631,107 @@ def select_groups(
...
+    outer_select = _select_sql(
+      outer_duckdb_path,
+      flatten=True,
+      unnest=True,
+      path=outer_path,
+      schema=manifest.data_schema,
+      span_from=self._resolve_span(outer_path, manifest),
+    )
+    filters, _ = self._normalize_filters(filters, col_aliases={}, udf_aliases={}, manifest=manifest)
+    if manifest.data_schema.has_field((DELETED_LABEL_NAME,)):
```
brilee, line 1702: next 8 lines could be gotten for free using

`where_query = self._compile_sql_options(DuckDBQueryParams(filters=filters, include_deleted=False))`

lilac/data/dataset_duckdb.py
```diff
@@ -1629,6 +1631,107 @@ def select_groups(
...
+      filters.append(Filter(path=(DELETED_LABEL_NAME,), op='not_exists'))
+    filter_queries = self._create_where(manifest, filters)
+
+    where_query = ''
+    if filter_queries:
+      where_query = f"WHERE {' AND '.join(filter_queries)}"
+
+    query = f"""
```
brilee, line 1710:
```
WITH double_group_by AS (
  SELECT {outer_select} AS out_val, {inner_select} AS in_val, count() AS c
  FROM t
  {where_query}
  GROUP BY out_val, in_val
)
SELECT
  out_val AS {value_column},
  SUM(c) AS {count_column},
  list({{'{value_column}': in_val, '{count_column}': c}} ORDER BY c DESC) AS {inner_column}
GROUP BY out_val
ORDER BY {sort_by.value} {sort_order.value}, {value_column}
```

brilee (approved):

lilac/data/clustering.py
dsmilkov: done

lilac/data/dataset.py
dsmilkov: done

lilac/router_dataset.py
dsmilkov: done

web/blueprint/src/lib/components/datasetView/DatasetPivotResult.svelte
dsmilkov: yes, due to faster rendering

lilac/data/clustering.py
dsmilkov: talked offline, sometimes the title is the name of a specific movie

lilac/data/dataset_duckdb.py
dsmilkov: the UI allows you to pivot by any two fields, even repeated ones, and duckdb auto handles the Cartesian product of (a, b) X (c, d, e) so it was magical that it just works

lilac/data/dataset.py
dsmilkov: done

lilac/data/dataset_duckdb.py
dsmilkov: done

PR #1075, opened 2024-01-17
Make the entire svelte app use relative links

dsmilkov:
Also in development serve the app both top level and under `/blueprint` so we can always test for regressions


web/blueprint/src/lib/components/datasetView/DatasetSettingsModal.svelte
```diff
@@ -197,7 +197,7 @@
                     onSuccess: () =>
                       // Invalidate the query after the redirect to avoid invalid queries to the
                       // dataset after it's deleted.
-                      goto('/').then(() => queryClient.invalidateQueries([DATASETS_TAG]))
+                      goto(startLink()).then(() => queryClient.invalidateQueries([DATASETS_TAG]))
```
nsthorat, line 200: homeLink

web/blueprint/src/routes/+layout.svelte
```diff
@@ -39,14 +40,17 @@
...
+  $: currentPage = routeToPage[$page.route.id || ''];
   let urlHashStore = createUrlHashStore(navStore);
   setUrlHashContext(urlHashStore);
 
   const notificationsStore = createNotificationsStore();
   setNotificationsContext(notificationsStore);
   $: notifications = $notificationsStore?.notifications || [];
 
+  // Set the base URL for OpenAPI requests automatically.
+  OpenAPI.BASE = base;
```
nsthorat, line 52: move this to the very top just in case we add stuff and get confused when it doesn't work?

web/blueprint/vite.config.ts
```diff
@@ -20,7 +23,12 @@ export default defineConfig({
       '/load_config': 'http://127.0.0.1:5432',
       // OpenAPI docs
       '^/docs': 'http://127.0.0.1:5432',
-      '/openapi.json': 'http://127.0.0.1:5432'
+      '/openapi.json': 'http://127.0.0.1:5432',
+      // Simulate the app being served under /blueprint base path for testing.
+      '^/blueprint': {
```
nsthorat, line 28: i can see this getting mega confusing, maybe this should be called "test_proxy"

nsthorat (approved):
nice work thanks for finishing it

web/blueprint/vite.config.ts
dsmilkov: done, called it /proxy

web/blueprint/src/routes/+layout.svelte
dsmilkov: done

PR #1082, opened 2024-01-17
Add remote computation for GTE

dsmilkov:


lilac/config.py
```diff
@@ -62,6 +62,10 @@ class EmbeddingConfig(BaseModel):
 
   path: PathTuple
   embedding: str
+  remote: bool = (
```
brilee, line 65: should probably default to False?

brilee (approved):

lilac/config.py
nsthorat: +1 here and elsewhere

lilac/data/dataset_duckdb.py
```diff
@@ -1030,7 +1030,8 @@ def compute_signal(
       self.project_dir,
     )
 
-    signal.setup()
+    if not remote:
```
nsthorat, line 1033: i think you should leave to the signal to decide

lilac/data/dataset_utils.py
```diff
@@ -41,9 +41,9 @@
 from ..utils import chunks, is_primitive, log, open_file
 
 # The embedding write chunk sizes keeps the memory pressure lower as we iteratively write to the
-# vector store. Embeddings are float32, taking up 4 bytes, so this results in ~130K * dims of RAM
+# vector store. Embeddings are float32, taking up 4 bytes, so this results in ~1.25MB * dims of RAM
 # pressure.
-EMBEDDINGS_WRITE_CHUNK_SIZE = 32_768
+EMBEDDINGS_WRITE_CHUNK_SIZE = 32_768 * 10
```
nsthorat, line 46: revert this or just type the number without multiplication?

lilac/embeddings/gte.py
```diff
@@ -32,6 +39,7 @@ class GTESmall(TextEmbeddingSignal):
   local_batch_size: ClassVar[int] = SENTENCE_TRANSFORMER_BATCH_SIZE
   local_parallelism: ClassVar[int] = 1
   local_strategy: ClassVar[TaskExecutionType] = 'threads'
+  runs_remote: ClassVar[bool] = True
```
nsthorat, line 42: - i think we should rename this to supports_garden for code marketing (and elsewhere, but maybe for now just do it here)

nsthorat (approved):

lilac/config.py
dsmilkov: done

lilac/data/dataset_utils.py
dsmilkov: done

lilac/data/dataset_duckdb.py
dsmilkov: done, added signal.setup_remote()

lilac/embeddings/gte.py
dsmilkov: done

PR #1085, opened 2024-01-18
Deploy dataset

brilee:
deploy_project(_operations) no longer does any dataset uploading; that is expected to be handled by the caller. Deploy is purely about uploading the project metadata and configuring the hf space.

There are four entry points to deploy_project_operations to double-check:
- CLI; not sure how this is used
- deploy_demo; this has an explicit project config at lilac_hf_space.yml that should be the source of truth, even if we only fiddle with one dataset at a time
- deploy_staging; this has no explicit project config and has to be dynamically constructed based on the dataset that gets uploaded
- deployer app; this constructs a project config and does no data upload (relies on the hf docker start's load() function to download from HF)


lilac/splitters/spacy_splitter.py
```diff
@@ -115,7 +115,7 @@ def _find_long_spans(breakpoints: list[int]) -> Optional[tuple[int, int]]:
 
 
 def clustering_spacy_chunker(
-  text: str, filter_short: int = 4, max_len: int = 512, target_num_groups: Optional[int] = None
+  text: str, filter_short: int = 0, max_len: int = 512, target_num_groups: Optional[int] = None
```
brilee, line 118: this was causing issues with some len<4 concept examples. Will revert before merge.

brilee (commented):
ready for review

lilac/cli.py
```diff
@@ -175,32 +176,39 @@ def deploy_project_command(
...
   hf_token = hf_token or env('HF_ACCESS_TOKEN')
+  ##
+  ##  Upload datasets.
+  ##
+  for d in dataset:
+    upload(
+      dataset=d,
+      project_dir=project_dir,
+      url_or_repo=get_hf_dataset_repo_id(*hf_space.split('/'), *d.split('/')),
+      public=False,
```
nsthorat, line 204: this should be a flag because we are going to make this True for the public demo

scripts/deploy_staging.py
```diff
@@ -107,17 +111,32 @@ def deploy_staging(
   if concept is None:
     concept = []
 
+  ##
+  ##  Upload datasets.
+  ##
+  if not skip_data_upload:
```
nsthorat, line 117: i think we can remove this bit since you can run this script without --dataset and it will just kick the demo with new code

nsthorat (commented):
Checked out your branch and getting some errors:
image

nsthorat (approved):

PR #1102, opened 2024-01-19
Move use_garden outside of a Signal.

nsthorat:
`use_garden` is now defined in configs, and manifests.

Details:
- Move `use_garden` in configs to the top-level so we don't define it everywhere. Now we only run on garden if a signal supports it, when the bit is true.
- Add UI for use_garden when a signal supports it, disabling it otherwise.

image

### DIFF
--- a/lilac/config.py
+++ b/lilac/config.py
@@ -62,9 +62,6 @@ class EmbeddingConfig(BaseModel):
 
   path: PathTuple
   embedding: str
-  use_garden: bool = PydanticField(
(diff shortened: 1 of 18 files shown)


dsmilkov:
Let's provide a bit more info next to the checkbox, like we do at head (screenshot below). Later we can link to the signup form in the subtext:

CleanShot 2024-01-19 at 19 48 55@2x

dsmilkov (approved):

nsthorat:
done!

PR #1103, opened 2024-01-20
Add accelerated PII execution on Lilac Garden

dsmilkov:
Also
- fix a bug with resolving spans that point to a string that was produced by a map output.
- fix a bug where the number of matching results is small that the requested page size, so there is no lookahead id.


lilac/data/dataset_duckdb.py
```diff
@@ -3292,48 +3279,91 @@ def _assert_embedding_exists(self, path: PathTuple, embedding: str) -> None:
     if not manifest.data_schema.has_field(embedding_path):
       raise ValueError(f'Embedding "{embedding}" not found for path {path}.')
 
+  def _inner_select(
```
dsmilkov, line 3282: this is just code move to get access to self

nsthorat (approved):

PR #1113, opened 2024-01-22
Export to HuggingFace. Support glaive-function-calling-v2 in the demo, clusters, and via sharegpt.

nsthorat:
Add `Dataset.to_huggingface` which returns a HuggingFace dataset. Users can then publish it explicitly.

To get the glaive-function-calling-v2 dataset to work we had to:
- Write a notebook to map() the data in lilac over to the sharegpt format.

Here is the output dataset: https://huggingface.co/datasets/lilacai/glaive-function-calling-v2-sharegpt

Demos:
- https://lilacai-lilac.hf.space/datasets#lilac/glaive-function-calling-v2
- https://lilacai-lilac.hf.space/datasets#lilac/glaive-function-calling-v2&viewPivot=true&pivot=%7B%22outerPath%22%3A%5B%22conversation_clusters%22%2C%22category_title%22%5D%2C%22innerPath%22%3A%5B%22conversation_clusters%22%2C%22cluster_title%22%5D%7D

Other:
- Allow this ./run_server_dev.sh ./demo_data

Input:
image

Output:
image

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -3188,6 +3188,7 @@ class DatasetDuckDB(Dataset):
       filter_likes=filters, col_aliases={}, udf_aliases={}, manifest=self.manifest()
     )
     filters.extend(self._compile_include_exclude_filters(include_labels, exclude_labels))
+    print('filters=', filters)
     rows = self.select_rows(
       columns, filters=filters, combine_columns=True, include_deleted=include_deleted
     )
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -3188,7 +3188,6 @@ class DatasetDuckDB(Dataset):
       filter_likes=filters, col_aliases={}, udf_aliases={}, manifest=self.manifest()
     )
     filters.extend(self._compile_include_exclude_filters(include_labels, exclude_labels))
-    print('filters=', filters)
     rows = self.select_rows(
       columns, filters=filters, combine_columns=True, include_deleted=include_deleted
     )
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -26,6 +26,7 @@ import numpy as np
 import orjson
 import pandas as pd
 import yaml
+from datasets import Dataset as HuggingFaceDataset
 from pandas.api.types import is_object_dtype
 from pydantic import BaseModel, SerializeAsAny, field_validator
 from typing_extensions import override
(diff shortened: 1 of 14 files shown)


lilac/data/dataset_duckdb.py
```diff
@@ -3174,6 +3175,30 @@ def cluster(
...
+    filters: Optional[Sequence[FilterLike]] = None,
+    include_labels: Optional[Sequence[str]] = None,
+    exclude_labels: Optional[Sequence[str]] = None,
+    include_deleted: bool = False,
+  ) -> HuggingFaceDataset:
+    filters, _ = self._normalize_filters(
+      filter_likes=filters, col_aliases={}, udf_aliases={}, manifest=self.manifest()
+    )
+    filters.extend(self._compile_include_exclude_filters(include_labels, exclude_labels))
+    print('filters=', filters)
```
dsmilkov, line 3191: remove print

dsmilkov (approved):

dsmilkov:
niceee!

lilac/db_manager.py
```diff
@@ -47,6 +47,21 @@ def get_dataset(
...
+) -> bool:
+  """Get the dataset instance."""
+  if not _DEFAULT_DATASET_CLS:
+    raise ValueError('Default dataset class not set.')
+
+  project_dir = project_dir or get_project_dir()
+  try:
+    get_dataset(namespace, dataset_name, project_dir)
+    return True
+  except ValueError:
```
brilee, line 61: not sure when this would ever fire? Looking at the body of get_dataset, it looks like the function should really be called "get_or_create_dataset"

brilee: the body of this function should probably be `return cache_key in _CACHED_DATASETS`

run_server_dev.sh
```diff
@@ -11,7 +11,11 @@ npm run dev --workspace web/blueprint -- --open &
 pid[2]=$!
 
 # Run the FastAPI server.
-export LILAC_PROJECT_DIR='./data'
```
brilee, line 14: oh, I ran into this myself - I was trying to solve it by looking up syntax for "export env variable unless it is already defined", but got distracted...

brilee (approved):

lilac/db_manager.py
nsthorat: lol it yells when the Dataset constructor cant find the manifest.json file

run_server_dev.sh
nsthorat: EXPORT IF NOT DEFINED OR SOMETHING OKAY?

PR #1121, opened 2024-01-24
Fix a few bugs related to concepts and clustering

dsmilkov:
https://huggingface.co/spaces/lilacai/daniel_staging (open capybara, conversations.*.input has embedding index)

- Fix a bug where deleting a concept creates 404s
- Fix a bug with old instructor code that generates positive examples
- Fix a render bug with concepts with 0 positive examples
- Fix a bug with clustering where the text contains the "__value__" substring, literally
- Fix a bug with lifecycle fastapi when server started in blocking mode
- Remove PALM which is timing out
- Add `EmbeddingInfo` to the field , so we can more easily find embeddings


web/blueprint/src/lib/queries/conceptQueries.ts
```diff
@@ -35,7 +37,11 @@ export const editConceptMetadataMutation = createApiMutation(ConceptsService.edi
 });
 
 export const deleteConceptMutation = createApiMutation(ConceptsService.deleteConcept, {
-  onSuccess: () => queryClient.invalidateQueries([CONCEPTS_TAG])
+  onSuccess: () => {
+    // Invalidate the query after the redirect to avoid invalid queries to the
+    // concept after it's deleted.
+    goto(homeLink()).then(() => queryClient.invalidateQueries([CONCEPTS_TAG]));
```
nsthorat, line 43: this feels like the wrong place for this, the redirect should be in the component that calls the mutation

i would bring this back to the original location

nsthorat (approved):

lilac/schema.py
```diff
@@ -255,6 +255,13 @@ class ClusterInfo(BaseModel):
   input_format_selector: Optional[ClusterInputFormatSelectorInfo] = None
 
 
+class EmbeddingInfo(BaseModel):
+  """Holds information about an embedding operation that was run on a dataset."""
+
+  input_path: Optional[PathTuple] = None
```
brilee, line 261: Does this make sense in the schema for Field? If embedding_info is an attribute of field, then input_path is already implicitly present by virtue of being a specific field object

brilee (approved):

web/blueprint/src/lib/queries/conceptQueries.ts
dsmilkov: nice catch. done

lilac/schema.py
dsmilkov: talked on slack

PR #1122, opened 2024-01-24
Fix concept labeler when we index a repeated string (capybara)

dsmilkov:
- Fix concept labeler for repeateds
- Make labeler fast again by caching the embedding models locally
- Make concept metrics fast by issuing 1 request instead of 7-8 requests (for each registered embedding)

### DIFF
--- a/web/blueprint/src/lib/components/concepts/labeler_utils.ts
+++ b/web/blueprint/src/lib/components/concepts/labeler_utils.ts
@@ -3,6 +3,7 @@ import {
   L,
   ROWID,
   valueAtPath,
+  valuesAtPath,
   type Concept,
   type DataTypeCasted,
   type LilacValueNode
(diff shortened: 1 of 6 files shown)


web/blueprint/src/lib/components/concepts/labeler_utils.ts
```diff
@@ -47,55 +48,53 @@ export function getCandidates(
...
-        if (span != null) {
-          labeledSpans.push(span);
-        }
-      }
-    }
-    for (const spanNode of spanNodes) {
-      const span = L.span(spanNode);
-      if (span == null) {
+    const textNodes = valuesAtPath(row, fieldPath);
+    for (const textNode of textNodes) {
```
dsmilkov, line 52: it's just an indented for loop (since repeated paths return many text nodes)

web/blueprint/src/lib/components/concepts/labeler_utils.ts
```diff
@@ -47,55 +48,53 @@ export function getCandidates(
...
+      ]) as unknown as LilacValueNode[];
+      if (spanNodes == null) {
         continue;
       }
-
-      const scoreNode = valueAtPath(spanNode, ['score']);
-      if (scoreNode == null) {
-        continue;
+      const labelNodes = valueAtPath(textNode, [
+        `${conceptId}/labels/preview`
```
dsmilkov, line 65: note had to append preview to the path (old path was outdated)

brilee (approved):

PR #1126, opened 2024-01-24
Cache dataset.pivot() and make cluster search box more visible

dsmilkov:

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -314,6 +314,9 @@ def _consume_iterator(iterator: Iterator[Any]) -> None:
     pass
 
 
+PivotCacheKey = tuple[PathTuple, PathTuple, GroupsSortBy, SortOrder]
+
+
 class DatasetDuckDB(Dataset):
   """The DuckDB implementation of the dataset database."""
 
@@ -344,6 +347,9 @@ class DatasetDuckDB(Dataset):
     self._config_lock = threading.Lock()
     self._vector_index_lock = threading.Lock()
     self._label_file_lock: dict[str, threading.Lock] = defaultdict(threading.Lock)
+
+    # Cache pivot results.
+    self._pivot_cache: dict[PivotCacheKey, PivotResult] = {}
 
     # Create a join table from all the parquet files.
     self.manifest()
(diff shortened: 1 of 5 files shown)


brilee (approved):

lilac/data/dataset_duckdb.py
```diff
@@ -1843,6 +1849,12 @@ def pivot(
     sort_order = sort_order or SortOrder.DESC
     inner_path = normalize_path(inner_path)
     outer_path = normalize_path(outer_path)
+
+    pivot_key = (outer_path, inner_path, sort_by, sort_order)
+    use_cache = not filters
+    if use_cache and pivot_key in self._pivot_cache:
```
nsthorat, line 1855: what happens when data is deleted?

also, does the pivot query remove deleted rows (it should)

web/blueprint/src/lib/components/datasetView/DatasetPivotViewer.svelte
```diff
@@ -162,15 +163,18 @@
 
 <div class="flex h-full flex-col">
   <div class="mb-8 flex h-16 w-full flex-row justify-between justify-items-center gap-x-4">
-    <div class="ml-8 mt-4 w-96">
-      <Search
+    <div class="ml-8 mt-4 flex w-96 items-center gap-x-2">
```
nsthorat, line 166: while you're here lets round the edges so it looks less IBM (and more like the searchbox and buttons above)

web/blueprint/src/lib/components/datasetView/DatasetPivotViewer.svelte
```diff
@@ -162,15 +163,18 @@
...
         value={searchText}
         on:input={searchInput}
-        placeholder={`Search`}
-        labelText={'text text'}
+        placeholder="Search all categories and titles"
         on:change={search}
-        on:clear={clearSearch}
+        size="xl"
       />
+      <div use:hoverTooltip={{text: 'Clear search query'}}>
```
nsthorat, line 175: make this invisible / disabled when theres nothing in  the searchbox

nsthorat (approved):
Can we achieve consistency with the top search box:

image

Also notice how google aligns, the edge of the content aligns with the edge of the searchbox, so you get a straight edge with text:
image

Also, it's really weird that one of these icons is a button and the other isnt, can we make the search a button (and just move it to the right, like Google):
image

web/blueprint/src/lib/components/datasetView/DatasetPivotViewer.svelte
dsmilkov: done

PR #1134, opened 2024-01-25
Rename USE_TABLE_INDEX => LILAC_USE_TABLE_INDEX. Add LILAC_PROD_MODE.

nsthorat:
Renames USE_TABLE_INDEX to LILAC_USE_TABLE_INDEX, and then makes sure it's backwards compat.

Add LILAC_PROD_MODE, which disables toast notification errors for the public demo, which makes it just look janky and broken.

Also add embeddings to missing datasets.

### DIFF
--- a/lilac/env.py
+++ b/lilac/env.py
@@ -44,7 +44,7 @@ class LilacEnvironment(BaseModel):
   LILAC_USE_TABLE_INDEX: str = PydanticField(
     description='Use persistent tables with rowid indexes.'
   )
-  LILAC_PROD_MODE: str = PydanticField(
+  LILAC_DISABLE_ERROR_NOTIFICATIONS: str = PydanticField(
     description='Set lilac in production mode. This will disable error messages in the UI.'
   )
 
--- a/lilac/env.py
+++ b/lilac/env.py
@@ -37,7 +37,16 @@ class LilacEnvironment(BaseModel):
     description='Turn on Lilac debug mode to log queries and timing information.'
   )
   DISABLE_LOGS: str = PydanticField(description='Disable log() statements to the console.')
-  USE_TABLE_INDEX: str = PydanticField(description='Use persistent tables with rowid indexes.')
+  USE_TABLE_INDEX: str = PydanticField(
+    description='Use persistent tables with rowid indexes.'
+    ' NOTE: This is deprecated in favor of USE_TABLE_INDEX.'
+  )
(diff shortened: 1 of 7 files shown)


lilac/env.py
```diff
@@ -37,7 +37,16 @@ class LilacEnvironment(BaseModel):
...
   DISABLE_LOGS: str = PydanticField(description='Disable log() statements to the console.')
-  USE_TABLE_INDEX: str = PydanticField(description='Use persistent tables with rowid indexes.')
+  USE_TABLE_INDEX: str = PydanticField(
+    description='Use persistent tables with rowid indexes.'
+    ' NOTE: This is deprecated in favor of USE_TABLE_INDEX.'
+  )
+  LILAC_USE_TABLE_INDEX: str = PydanticField(
+    description='Use persistent tables with rowid indexes.'
+  )
+  LILAC_PROD_MODE: str = PydanticField(
```
dsmilkov, line 47: LILAC_DISABLE_TASK_NOTIFICATIONS ? 

Prod mode will probably mean something else in 6 months from now

dsmilkov (approved):

lilac/env.py
nsthorat: renamed to LILAC_DISABLE_ERROR_NOTIFICATIONS

web/blueprint/src/lib/components/ErrorNotifications.svelte
```diff
@@ -0,0 +1,33 @@
...
+  import type {ApiError} from '$lilac';
+  import ApiErrorModal from './ApiErrorModal.svelte';
+
+  let showError: ApiError | undefined = undefined;
+
+  $: serverStatus = queryServerStatus();
+  $: disableErrorNotifications = $serverStatus.data?.disable_error_notifications;
+</script>
+
+{#if !disableErrorNotifications}
```
brilee, line 15: Do we have a way to see prod errors regardless? I guess they would be in the HF space logs. Maybe we should set up some sort of error tracker (probably a good idea for learning about the mechanics of hosting others' servers, in any case...)

brilee (approved):

web/blueprint/src/lib/components/ErrorNotifications.svelte
nsthorat: They'll show up in server side logs and console logs. I'll basically only set this bit for the shared demo. Users won't set this, most likely.

PR #1139, opened 2024-01-26
Add exclude_signals to select_rows, and include_signals to export methods.

nsthorat:
The motivation here is for export methods to quickly allow disabling signals for exporting.

- `select_rows` now has `exclude_signals`, which automatically disables signals.
- `to_*` now have `include_signals`, which by default removes signals. If you set this to true, you will get signals on the other side.

Also add a cluster sampling notebook that shows this workflow.

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -1,4 +1,5 @@
 """The DuckDB implementation of the dataset database."""
+import copy
 import csv
 import functools
 import gc
@@ -2051,6 +2052,7 @@ class DatasetDuckDB(Dataset):
     resolve_span: bool = False,
     combine_columns: bool = False,
     include_deleted: bool = False,
+    exclude_signals: bool = False,
     user: Optional[UserInfo] = None,
   ) -> SelectRowsResult:
     manifest = self.manifest()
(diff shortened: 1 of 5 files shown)


lilac/data/dataset_duckdb.py
```diff
@@ -2640,7 +2666,7 @@ def _column_to_duckdb_paths(
     is_petal = schema.get_field(path).dtype is not None
 
     # NOTE: The order of this array matters as we check the source and map manifests for fields
-    # before reading signal manifests, via source_or_map_has_path.
+    # before reading signal manifests, via source_or_map_has_path.s
```
dsmilkov, line 2669: typo

lilac/data/dataset_test.py
```diff
@@ -283,6 +283,27 @@ def _get_rows() -> list[Item]:
...
+
+  # Compute a map and make sure it shows up.
+  dataset.map(lambda row: row['text'] + '_mapped', output_path='text_map')
+
+  test_signal = TestSignal()
+  dataset.compute_signal(test_signal, 'text')
+  length_signal = LengthSignal()
+  dataset.compute_signal(length_signal, 'text')
+
+  result = dataset.select_rows(combine_columns=True, exclude_signals=True)
```
dsmilkov, line 297: can you add a test where you select explicitly just a single sub-field of a signal (via columns, but exclude_signals=True)

dsmilkov (approved):

lilac/data/dataset_test.py
nsthorat: done

lilac/data/dataset_duckdb.py
nsthorat: done

PR #1140, opened 2024-01-26
Handle sqlite files separately during table/index creation 

brilee:
Sqlite files are now not included in the cached table; this means that label updates are instantaneous as they would be with LILAC_USE_TABLE_INDEX=false.

Some resulting changes:
- sqlite files are considered separately in cachebusting logic
- type of `t` switches from TABLE to VIEW, causing various duckdb errors. As a result...
- new generic try-catch to handle all sorts of migration/transition errors.

Users may see errors in their logs like

```
Exception: Catalog Error: Table with name cache_t does not exist!
Did you mean "t"?
Exception encountered while updating joint table cache; recomputing from scratch.
Recomputing table+index for $DATASET... took 0.154s.
```

or 
```
Catalog Error: t is not an table
```

### DIFF
--- a/lilac/data/dataset_duckdb.py
+++ b/lilac/data/dataset_duckdb.py
@@ -439,9 +439,14 @@ class DatasetDuckDB(Dataset):
       JOIN labels USING (rowid)
       JOIN maps USING (rowid))
 
-    Signals and maps are asynchronous, slow requests that act on the entire table.
-    Labels are synchronous fast requests that insert single rows into a sqlite table.
-    In the scenario where we are using cached Tables/indices, we separate out sqlite tables.
+    If LILAC_USE_TABLE_INDEX is set, then we apply a further optimization where we create a DuckDB
+    table as a cache and index over Sources, Signals, and Maps (excluding labels). We invalidate
+    this cache if the underlying data is updated. Labels are excluded from the cache because users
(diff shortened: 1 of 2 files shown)


lilac/data/dataset_duckdb.py
```diff
@@ -531,13 +556,63 @@ def _recompute_joint_table(self, latest_mtime_micro_sec: int) -> DatasetManifest
...
+          # What if cache_t exists but t does not?
+          # What if the type of t was VIEW but we replace it with a TABLE or vice-versa?
+          # What if the user deletes the duckdb.cache while server is up?
           self.con.execute('UPDATE mtime_cache SET mtime = ?', (latest_mtime_micro_sec,))
-          self.con.execute(f'CREATE OR REPLACE TABLE t AS (SELECT {select_sql} FROM {join_sql})')
-          self.con.execute('CREATE INDEX row_idx ON t ("__rowid__")')
+          self.con.execute(
+            f'CREATE OR REPLACE TABLE cache_t AS (SELECT {table_select_sql} FROM {table_join_sql})'
+          )
+          self.con.execute('CREATE INDEX row_idx ON cache_t ("__rowid__")')
```
nsthorat, line 592: nit ROWID const

nsthorat (approved):

PR #1141, opened 2024-01-29
Improve clusters and several bug fixes

dsmilkov:
Clustering (backend):
- Lower `min_cluster_size` to 5 for the categories, so we can have more coherent categories
- Add timeout of 7 sec (99%-tile response latency for OpenAI is like 3-4 sec) to avoid the 10min timeout
- Disable internal OpenAI retries (we used to have double retries)
- Replace "request" with "snippet" in the prompt to avoid biasing towards user's requests -- improves forum/email/text clustering

UI
- make the histograms reactive to the currently selected group in "group by"
- make pivot reactive to searches (e.g. keyword search, metadata search)
- remember the schema and nav bar state when flipping between cluster view and item view
- Fix a bug with search box state, after page refresh

### DIFF
--- a/lilac/data/clustering.py
+++ b/lilac/data/clustering.py
@@ -44,7 +44,7 @@ from .dataset_utils import (
 
 _SHORTEN_LEN = 400
 _TOP_K_CENTRAL_DOCS = 7
-_TOP_K_CENTRAL_TITLES = 15
+_TOP_K_CENTRAL_TITLES = 20
 _NUM_THREADS = 32
 
 CLUSTER_ID = 'cluster_id'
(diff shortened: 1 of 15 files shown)


lilac/data/clustering.py
```diff
@@ -44,8 +44,11 @@
 
 _SHORTEN_LEN = 400
 _TOP_K_CENTRAL_DOCS = 7
-_TOP_K_CENTRAL_TITLES = 15
+_TOP_K_CENTRAL_TITLES = 20
 _NUM_THREADS = 32
+_NUM_RETRIES = 16
+_INITIAL_MAX_TOKENS = 50
```
nsthorat, line 50: might be worth a comment about what these two mean

lilac/data/clustering.py
```diff
@@ -167,29 +184,45 @@ def generate_category(ranked_docs: list[tuple[str, float]]) -> str:
...
+            },
+            {'role': 'user', 'content': input},
+          ],
+        )
+        return category.category
+      except IncompleteOutputException:
+        max_tokens += _INITIAL_MAX_TOKENS
+        log(f'Retrying with max_tokens={max_tokens}')
+    log(f'Could not generate a short category for input:\n{input}')
+    return 'FAILED_TO_GENERATE'
```
nsthorat, line 225: should this be a None?

brilee (approved):

nsthorat (approved):

PR #1151, opened 2024-01-31
Fix unusual inputs to auto binning histogram

brilee:


nsthorat (approved):

brilee:
Screenshot 2024-01-31 at 2 15 50 PM

how it renders in the UI

PR #1159, opened 2024-02-01
Fix issue where by default a long media doesnt take up the full screen

nsthorat:
https://lilacai-nikhil-staging.hf.space/datasets#local/mosaic-chat-v2&rowId=%22000006e20ceb477f889e297c22f4e2f2%22

image


nsthorat:
This is done, but github does not like to load on airplanes.

Issue #1170, opened 2024-02-07
Mount Lilac in existing FastAPI App


derek-pyne:
Love Lilac, looking to mount it in our existing FastAPI setup for easy deploy (we've got other APIs and Gradios Apps already in there).

This would be a nice way to fit Lilac into the other suite of tools folks are using to build their own little llm gardens :)

Don't want to use `lilac.start_server` as the spins up the Uvicorn Server. Since the goal is to mount within an existing FastAPI App, this is a duplicate Uvicorn.

There is a FastAPI App within `server.app`, but, after mounting, responses are infinite 307s (`HttpUrlRedirectMiddleware` probably isn't meant to be mounted like this)

Interesting.... the docs work great! (ie. `lilac/docs` in the example)

Minimal example:

```python
import lilac.server
import uvicorn
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
def read_main():
    return {"message": "Hello World from main app"}


app.mount("/lilac", lilac.server.app)

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8000)

```

PR #1174, opened 2024-02-13
Support the FastAPI app being mounted.

nsthorat:
Fixes https://github.com/lilacai/lilac/issues/1170


dsmilkov (approved):

integration_tests/fastapi_mount_test.sh
```diff
@@ -0,0 +1,29 @@
...
+    echo "Timeout $TIMEOUT_SEC seconds to reach server."
+    kill $pid
+    exit 1
+  fi
+done
+
+
+curl --fail --silent "http://localhost:$PORT/lilac_sub/api/v1/tasks/"
+
+curl --fail --silent "http://localhost:$PORT/lilac_sub/"
```
derek-pyne, line 29: The issue we're seeing with mounting lilac is the infinite redirects... and this test isn't quite catching it right now. Changing to this would show the failure:

`curl --fail --silent "http://localhost:$PORT/lilac_sub/" -L --max-redirs 0`

and if you remove `--silent` show something like this:
Screenshot 2024-02-14 at 10 50 44 AM

Interesting that the api and docs work totally fine :)

nsthorat:
I haven't yet released this! Did you try checking out from head?

I fixed the infinite redirect, but good point I will fix the unit test :)

Issue #1148, opened 2024-01-31
can we add this open source multilingual embedding model?


kostum123:
It would be nice if we add this model as a multilingual embedding model.

https://huggingface.co/BAAI/bge-m3

kostum123:
> Happy to do this -- is there a particular size you care most about?

There is currently only one version for the multilingual version, it has 568M parameters.  It has 3 modes in 1 model: dense retrieval, multi-vector(colbert like) retrieval, and sparse retrieval.  It seems to be the best among open source multilingual models.  It can produce embedding with a combination of 3 modes and has a context length of 8k.  I think Lilac's addition of such a model would be an alternative for users who want to use it in multilingual datasets.

![imgs_miracl](https://github.com/lilacai/lilac/assets/34637927/968dbdaf-b465-41f9-a60b-d9f5363c73f7)
![imgs_mkqa](https://github.com/lilacai/lilac/assets/34637927/91b0b38e-8ae0-4f16-82ac-baa23d6b8f45)

nsthorat:
bge-m3 is supported in https://github.com/lilacai/lilac/releases/tag/v0.3.5

PR #1182, opened 2024-02-14
Add bge-m3 and Nomic 1.5 embeddings.

nsthorat:
Fixes https://github.com/lilacai/lilac/issues/1148

I added two versions of the Nomic 1.5 embeddings, which use matryoshka_dim dims, see: https://huggingface.co/nomic-ai/nomic-embed-text-v1.5

Demo: https://lilacai-nikhil-staging.hf.space/datasets#local/glue_ax_map&schemaCollapsed=false


pyproject.toml
```diff
@@ -86,6 +89,7 @@ hdbscan = { version = "^0.8.33", optional = true }
 
 [tool.poetry.extras]
 all = [
+  "bge",
```
dsmilkov, line 92: these have to be pypi packages, not aliases to groups

lilac/embeddings/nomic_embed.py
```diff
@@ -0,0 +1,109 @@
...
+      raise ImportError(
+        'Could not import the "sentence_transformers" python package. '
+        'Please install it with `pip install "sentence_transformers".'
+      )
+
+    def _encode(doc: list[str]) -> list[np.ndarray]:
+      embeddings = self._model.encode(doc, convert_to_tensor=True)
+      # Extract the dense vectors from the model.
+      embeddings = F.layer_norm(embeddings, normalized_shape=(embeddings.shape[1],))
+      embeddings = embeddings[:, : self._matryoshka_dim]
```
dsmilkov, line 73: are dimensions sorted by importance , like PCA ? curious why we can just take the last 256/512 dimensions

dsmilkov (approved):
nice!

lilac/embeddings/nomic_embed.py
nsthorat: thats the magic of matryoshka_dim.. it's not PCA exactly, afaict it's baked into the loss function: https://aniketrege.github.io/blog/2024/mrl/

pyproject.toml
nsthorat: done

PR #1183, opened 2024-02-14
Fix issue with FastAPI mounts.

nsthorat:


lilac/server.py
```diff
@@ -98,7 +98,11 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
     url = URL(scope=scope).path
     root_path = scope.get('root_path') or ''
     ends_with_slash = (
-      url.endswith('/') and url != '/' and root_path and not url.startswith(root_path + '/api')
```
brilee, line 101: can we make this a unit test to enumerate the kinds of URLs that this does/doesn't accept?

brilee (approved):

dsmilkov (approved):

integration_tests/fastapi_mount_test.sh
```diff
@@ -24,6 +24,5 @@ until curl --fail --silent "$URL" > /dev/null; do
 done
 
 
-curl --fail --silent "http://localhost:$PORT/lilac_sub/api/v1/tasks/"
-
-curl --fail --silent "http://localhost:$PORT/lilac_sub/"
+curl --fail --silent "http://localhost:$PORT/$MOUNT_PATH/api/v1/tasks/" > /dev/null
+curl --fail --silent "http://localhost:$PORT/$MOUNT_PATH/" --max-redirs 0 > /dev/null
```
derek-pyne, line 28:
```suggestion
curl --fail --silent "http://localhost:$PORT/$MOUNT_PATH/" -L --max-redirs 0 > /dev/null
```

Need to add `-L` (follow redirects) for `--max-redirs` to have any effect

lilac/server.py
nsthorat: done

integration_tests/fastapi_mount_test.sh
nsthorat: Thanks! Done :)

nsthorat: I had to migrate this to a python test, FYI.

PR #1193, opened 2024-02-27
Respect the `self._split` param when computing embeddings for a text

dsmilkov:
- When we sync a concept database with a concept model, we need to compute embeddings without chunking.
- Fix occasional `undefined is not iterable` js bug in the `ItemMediaTextContent`
- Remove "apply to dataset" in concept view, since the labeler has a "go to dataset" button

### DIFF
--- a/web/blueprint/src/lib/components/concepts/ConceptView.svelte
+++ b/web/blueprint/src/lib/components/concepts/ConceptView.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import {goto} from '$app/navigation';
   import {
     editConceptMutation,
     queryConceptModels,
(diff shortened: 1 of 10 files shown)


web/blueprint/src/lib/components/concepts/ConceptView.svelte
```diff
@@ -45,25 +38,6 @@
 
   $: randomPositive = positiveExamples[Math.floor(Math.random() * positiveExamples.length)];
 
-  // Apply to a dataset.
```
nsthorat, line 48: revert this? you're not always doing this via collect labels

nsthorat (approved):

PR #1194, opened 2024-02-28
Add `skip_noisy_assignment` to `dataset.cluster`

dsmilkov:
This reduces an expensive step for assigning noisy clusters. It's only effectively when running locally.

### DIFF
--- a/lilac/data/clustering.py
+++ b/lilac/data/clustering.py
@@ -53,7 +53,7 @@ MIN_CLUSTER_SIZE_CATEGORY = 5
 UMAP_DIM = 5
 UMAP_SEED = 42
 HDBSCAN_SELECTION_EPS = 0.05
-BATCH_SOFT_CLUSTER_NOISE = 1024
+BATCH_SOFT_CLUSTER_NOISE = 512
 
 
 def cluster_impl(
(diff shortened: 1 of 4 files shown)


lilac/data/clustering.py
```diff
@@ -68,6 +68,7 @@ def cluster_impl(
   task_id: Optional[TaskId] = None,
   recompute_titles: bool = False,
   batch_size_titling: Optional[int] = None,
+  skip_noisy_assignment: bool = False,
```
nsthorat, line 71: if garden and skip_noisy_assignment => throw error?

lilac/data/dataset.py
```diff
@@ -506,6 +506,7 @@ def cluster(
     task_id: Optional[TaskId] = None,
     # TODO(0.4.0): colocate with topic_fn.
     category_fn: Optional[TopicFn] = None,
+    skip_noisy_assignment: bool = False,
```
nsthorat, line 509: can we pipe this to the UI?

nsthorat (approved):
woop

lilac/data/clustering.py
dsmilkov: done

PR #1195, opened 2024-02-29
Fix a bug with excess RAM usage during vector computes.

nsthorat:
Vector stores now return Iterators of np.array vectors.

### DIFF
--- a/lilac/embeddings/vector_store.py
+++ b/lilac/embeddings/vector_store.py
@@ -159,13 +159,10 @@ class VectorDBIndex:
       all_spans.append(spans)
       all_vector_keys.append([(*path_key, i) for i in range(len(spans))])
 
-    offset = 0
     flat_vector_keys = [key for vector_keys in all_vector_keys for key in (vector_keys or [])]
-    all_vectors = self._vector_store.get(flat_vector_keys)
+    all_vectors = iter(self._vector_store.get(flat_vector_keys))
     for spans in all_spans:
-      vectors = all_vectors[offset : offset + len(spans)]
-      yield [{'span': span, 'vector': vector} for span, vector in zip(spans, vectors)]
-      offset += len(spans)
+      yield [{'span': span, 'vector': next(all_vectors)} for span in spans]
 
   def topk(
     self, query: np.ndarray, k: int, rowids: Optional[Iterable[str]] = None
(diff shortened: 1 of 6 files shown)


lilac/embeddings/vector_store.py
```diff
@@ -159,13 +159,10 @@ def get(self, keys: Iterable[PathKey]) -> Iterable[list[SpanVector]]:
       all_spans.append(spans)
       all_vector_keys.append([(*path_key, i) for i in range(len(spans))])
 
-    offset = 0
     flat_vector_keys = [key for vector_keys in all_vector_keys for key in (vector_keys or [])]
-    all_vectors = self._vector_store.get(flat_vector_keys)
+    all_vectors = iter(self._vector_store.get(flat_vector_keys))
```
brilee, line 163: don't think you need the iter?

lilac/embeddings/vector_store_hnsw.py
```diff
@@ -105,15 +107,20 @@ def add(self, keys: list[VectorKey], embeddings: np.ndarray) -> None:
...
-        return np.array(self._index.get_items(self._key_to_label.values), dtype=np.float32)
-      locs = self._key_to_label.loc[cast(list[str], keys)].values
-      return np.array(self._index.get_items(locs), dtype=np.float32)
+        locs = self._key_to_label.values
+      else:
+        locs = self._key_to_label.loc[cast(list[str], keys)].values
+
+      for loc_chunk in chunks(locs, HNSW_RETRIEVAL_BATCH_SIZE):
+        chunk_items = np.array(self._index.get_items(loc_chunk), dtype=np.float32)
+        for vector in np.split(chunk_items, chunk_items.shape[0]):
```
brilee, line 122: probably simpler/shorter:
```
for i in range(chunk_items.shape[0]):
  yield chunk_items[i]
```

brilee (approved):

lilac/embeddings/vector_store.py
nsthorat: yep, no longer needed. was using islice but simplified it.

lilac/embeddings/vector_store_hnsw.py
nsthorat: need to use split here since it doesn't make a copy

dsmilkov (approved):

